PR #41, opened 2020-01-20
Change mariadb docker command

flut1:
This is necessary to read database files from the host machine on Windows. 

For more information see:
https://github.com/docker-library/mariadb/issues/95#issuecomment-407839431

This seems to fix the issues with running the site locally on Windows.


flut1:
Someone with a mac, please verify this is not breaking things

PR #42, opened 2020-01-29
Changed the instagram link in header and footer to point to the extio…

franei:
Changed the instagram link in header and footer to point to the extionctionrebellionnl account (instead of extionctionrebellion).


SamLubbers:
@franei congratulations with your first contribution to the XR NL Website 🎉

PR #43, opened 2020-02-04
Removing navigation list headings in footer

franei:
This removes the headings above the two navigation lists in the page footer. The links are to be restructured into two lists with five items each, and the headings will no longer be appropriate.

Current footer: 

old_footer

New footer with the re-ordered navigation lists (which still needs to be done in WP admin once this is live):

new_footer

### DIFF
--- a/web/app/themes/xrnl/footer.php
+++ b/web/app/themes/xrnl/footer.php
@@ -7,7 +7,6 @@
                 </div>
 
                 <div class="col-md-4 my-4">
-                    <h3 class="font-xr text-uppercase">Extinction Rebellion</h3>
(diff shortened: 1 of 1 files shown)


franei:
Sure, this was just meant to implement the todo item from trello :) Agreed that headings would look good but it's not obvious to me what they should be.. 'Join' maybe doesn't quite fit with Blog, Press & Media, and FAQ? The best grouping I can come up with is 'About', 'Join', and 'Read more', if we wanted to make it three.

SamLubbers:
Just merged the branch so we can have the updated footers on the website. If we need new headers we can add them at a later point.

PR #47, opened 2020-03-05
Improved roles overview and details page

SamLubbers:
### Before
image
### After
image

### Before
image

### After
image


madhums:
This is great @SamLubbers! Thanks for taking the time and effort to implement this! 

Just one question: Do we have the content for the volunteering positions?

SamLubbers:
@madhums We will very very soon. 

Valentin from the integration circle will submit this weekend 5 roles for the integration circle, a few for A&L Amsterdam and I will submit some from tech.

Once we have some roles there it will be easier to promote 👌

web/app/themes/xrnl/single-volunteer_vacancy.php
```diff
@@ -0,0 +1,77 @@
...
+            <a href="mailto:<?php echo $role->email ?>?subject=Role application: <?php echo $role->title ?>" target="_blank">
+            <?php echo $role->email ?>
+            </a>
+        </div>
+        <div>
+            <i class="fas fa-comment mr-2"></i>
+            <b>Mattermost: </b><a href="https://organise.earth/xr-netherlands/messages/<?php echo $role->mattermostId ?>" target="_blank">
+            <?php echo $role->mattermostId ?>
+            </a>
+        </div>
```
madhums: Don't think we should be using mattermost links on public website, could you remove this?

web/app/themes/xrnl/volunteer_vacancy-archive.php
```diff
@@ -18,23 +18,28 @@
...
-          <?php } wp_reset_query(); ?>
-        </ol>
-
     </div>
   </div>
-
-  <div class="row my-4">
-    <div class="col-12 col-lg-8">
-      <?php the_field('how_to_volunteer') ?>
+        <div class="d-flex flex-wrap justify-content-center mt-3">
```
madhums: remove `justify-content-center`, left justification looks better.

madhums: Why not use something like [card groups](https://getbootstrap.com/docs/4.4/components/card/#card-groups) and [sizing utilities](https://getbootstrap.com/docs/4.4/components/card/#using-utilities) instead of `d-flex`, custom css and widths?

SamLubbers: Good suggestion! I'll soon implement a better alignment of cards.

web/app/themes/xrnl/single-volunteer_vacancy.php
SamLubbers: I will reply privately to this.

PR #50, opened 2020-04-10
hide action network api key

SamLubbers:
We want to load the API key from an `.env` file instead of the wordpress panel so that someone with access to the wordpress (a lot of people) cannot steal our API key. 

The only thing I have changed is line `730` of the `actionnetwork.php` file. We have concluded that it is okay to change the plugin code because this plugin isn't being maintained any more. 

If you think we shouldn't commit this file, I can also remove it and share instructions for what line of code needs to be changed on the server. 

When this PR is deployed, make sure to add the action network api key to the `.env` file. See `.env.example` for an example.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,3 +23,5 @@ AUTH_SALT='generateme'
 SECURE_AUTH_SALT='generateme'
 LOGGED_IN_SALT='generateme'
 NONCE_SALT='generateme'
+ACTION_NETWORK_API_KEY=INCLUDE_API_KEY_HERE
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 # Application
 web/app/plugins/*
 !web/app/plugins/.gitkeep
+# Excluded for custom loading of API key from .env file instead of wordpress admin
+!web/app/plugins/wp-action-network/
+web/app/plugins/wp-action-network/*
+!web/app/plugins/wp-action-network/actionnetwork.php
 web/app/mu-plugins/*/
 web/app/upgrade
 web/app/uploads/*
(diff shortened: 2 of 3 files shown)


SamLubbers:
@DaanMeijer if there is an option that does not require the plugin let's go for it. 

I don't understand your proposed solution. Could you explain in some more detail how I could implement it? from where do we write to the $_REQUEST array, and how does the plugin get the API_KEY without inserting it in the admin panel?

DaanMeijer:
This should be done in functions.php (or an included file). The current "active" code is:
```
switch ($_REQUEST['actionnetwork_admin_action']) {

	case 'update_api_key':

		$debug = "update_api_key case matched\n";

		// $actionnetwork_api_key = sanitize_key($_REQUEST['actionnetwork_api_key']);
		// removed default api mechanism to hide the api key form the wordpress panel. Loaded from .env instead.
		$actionnetwork_api_key = env('ACTION_NETWORK_API_KEY');
```

So something like:
```
if(@$_REQUEST['actionnetwork_admin_action'] == 'update_api_key'){
	$_REQUEST['actionnetwork_api_key'] = env('ACTION_NETWORK_API_KEY');
}
```

SamLubbers:
@DaanMeijer thank you for your suggestion. I tried implementing your solution but I am still facing a barrier due to my lack of experience with PHP.

From what I have learned in the past hours, I must associate the function that you suggested to a particular action in WordPress, in this case the action is submitting the API key from the Action Network admin panel. If I am correct, the way to do that is with `add_action`. The issue that I am facing is that I can't seem to figure to which action I must associate the code. 

See my code below. What I think I am missing is the right value in place of `?`

```php
// functions.php
add_action('?', function(){
	if($_REQUEST['actionnetwork_admin_action'] == 'update_api_key'){
		$actionnetwork_api_key = env('ACTION_NETWORK_API_KEY');
		$_REQUEST['actionnetwork_api_key'] = $actionnetwork_api_key;
	}
}, 1);
```

PR #53, opened 2020-04-20
Community group pages

franei:
This adds
- a template for single community group pages
- a template for an overview of community groups
- a new "Community group" post type in functions.php
- custom fields
- some CSS

families_page_hero_light
gemeenschapsgroepen_wide
custom_fields

### DIFF
--- a/web/app/themes/xrnl/community_group-archive.php
+++ b/web/app/themes/xrnl/community_group-archive.php
@@ -1,0 +1,69 @@
+<?php
+/*
+* Template name: Community groups
+*/
+
+get_header(); ?>
+
+<?php
+  $groups = new WP_Query([
+    'order' => 'DESC',
+    'orderby' => 'ID',
+    'post_type' => 'community_group',
+    'posts_per_page' => -1
+  ]);
+  $default_img_url = 'https://extinctionrebellion.nl/app/uploads/2020/04/community-group.png';
+?>
+
+<div id="community" class="container py-5">
+    <h1 class="mt-5"><?php the_title(); ?></h1>
+
+    <div class="row">
+      <?php while($groups->have_posts()):$groups->the_post(); ?>
+        <?php $show_link_to_group_page = !empty(get_the_content()); // Show link if group has content text ?>
+
+        <div class="col-8 col-sm-6 col-md-4 col-xl-4 mt-4 group-column">
+          <div id="group-card-<?php echo($groups->current_post); ?>" class="card group-card h-100 border-0">
+
+            <?php $img_url = get_field('group_cover_image_url') ?: $default_img_url;  // Use default pic if cover image is missing ?>
+            <img src="<?php echo($img_url); ?>" class="card-img-top" alt="XRNL Community group">
+
+             <?php if($show_link_to_group_page): ?>
+               <a href="<?php the_permalink(); ?>"><h5 class="card-header font-xr text-uppercase bg-yellow border-0"><?php the_title(); ?></h5></a>
+             <?php else :?>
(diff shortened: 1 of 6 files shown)


web/app/themes/xrnl/single-community_group.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * The template for a community group page
+ */
+
+get_header(); ?>
+
+<?php
+$communityPage = $translated_page = icl_object_id(6832, 'page', true); // XRNL: 6832 NL, 6853 EN
```
SamLubbers, line 9: I suspect that this is the ID of the communities page on your local instance of the website. When this is deployed on the live website it will not work as the community page created there will likely have another ID.

To fix this, you can create now the `/community` page on the live website and update this ID with the ID of the communities page of the live website. 

When this PR is deployed, you can configure the `/community` page to use the `community groups` template.

web/app/themes/xrnl/community_group-archive.php
```diff
@@ -0,0 +1,69 @@
...
+get_header(); ?>
+
+<?php
+  $groups = new WP_Query([
+    'order' => 'DESC',
+    'orderby' => 'ID',
+    'post_type' => 'community_group',
+    'posts_per_page' => -1
+  ]);
+  $default_img_url = 'https://extinctionrebellion.nl/app/uploads/2020/04/community-group.png';
```
SamLubbers, line 15: Ideally it should be possible to configure this image from the Wordpress panel. It's not a problem if we can't.

franei: Right, this is probably unnecessary because the cover image field in the WP panel already has the same URL as its default value. I was just thinking that a user could remove the default, and then this here would be the fallback.

web/app/themes/xrnl/single-community_group.php
franei: I'm quite surprised myself that I was thinking ahead that far, but I've actually already done this :) So these IDs are pages on the live site, one for the Dutch and one for the English version. As far as I could see, it doesn't matter which one is used.

SamLubbers:
### Translation of community post type
Here is a clarification of what I meant by translating community groups.

When you view a vacancy (or any other post), there are options on the navigation bar for changing the language of the page

Screenshot 2020-05-02 at 11 20 24

But for community groups this option does not exist.

Screenshot 2020-05-02 at 11 25 06

To make it possible to translate the text of community groups, there should be a `Translate` panel on the right side of the page to create/edit a community group. Here is a screenshot taken from the `vacancy` post type. 

Screenshot 2020-05-02 at 11 19 27

franei:
Strangely, this is working on my local setup. The language switcher is in the page header, and the Language tab is there in the admin. My best guess is that "Do not make 'Community Groups' translatable" is checked at the bottom of the admin page where you edit the post type.
It should be "Make 'Community Groups' translatable". Could you maybe try that (and click Apply) ?For me this is the default, not sure where you can explicitly set it.
community_groups_multilang

SamLubbers:
@franei You are right, all I had to do was to activate that option for the translations panel to appear. Sorry I didn't fix the problem myself, I assumed it was a problem with how the custom post type was configured.

### Link to community group pages
I have one more suggestion for this pull request before merging. It would be good to have a link to the communities page from somewhere else in the application. We could start by linking to it from the Join page. Once we have a lot of community groups we could also add it in the navigation bar or somewhere else more prominent.

web/app/themes/xrnl/single-community_group.php
SamLubbers: @franei that's amazing! I tried searching for `community` on the live site but couldn't find it. Now I realise that is because it is called `Gemeenschapsgroepen` 😌

PR #60, opened 2020-05-24
Redesign of volunteer pages

SamLubbers:
The roles overview page and the individual role page has been redesigned in accordance with the new [global XR page](https://vue.dev.rebellion.global/).

In addition to providing a better user experience, these changes also include some missing translations, better accessibility, and a responsive design which makes it easier to navigate on mobile.

#### Before
Screenshot_2020-05-24 De rebellie heeft jouw hulp nodig · Extinction Rebellion Netherlands(1)

#### After

Screenshot_2020-05-24 De rebellie heeft jouw hulp nodig · Extinction Rebellion Netherlands

#### Before
Screenshot 2020-05-24 at 15 53 48

#### After
Screenshot 2020-05-24 at 15 54 03

#### Before
Screenshot 2020-05-24 at 16 04 34

#### After
Screenshot 2020-05-24 at 16 04 56

### Before
Screenshot_2020-05-24 Front-end developer · Extinction Rebellion Netherlands(1)

### After
Screenshot_2020-05-24 Front-end developer · Extinction Rebellion Netherlands


SamLubbers:
Thanks for the review Frank. I've added the en-dashes to please the typography gods. Good observation about the special character not working, I have made issue #61 so we don't forget about it. 

And we should totally use those SVG more! In the future we can do a big redesign of the website so it looks more like the global website.

PR #66, opened 2020-06-18
New join page template and CFs

franei:
### New Join page layout
This PR implements a [new layout](https://organise.earth/xr-netherlands/pl/gjhwy6h84pbcidkkfwb7qejjto) for the Join page with a new template and custom field set. 

The main change from the previous version is that every section can be hidden from the admin panel with a switch:

section_switch

There are some new flexible elements in the groups, actions, and do-more sections.

share_btns

add_actions

This template has a new name because I'm thinking that this will make it easier to update the page without breaking it. This way we can take our time with setting up the new version and when it's ready we just change the permalink. Please tell me if this is bad idea :)

### DIFF
--- a/web/app/themes/xrnl/doe-mee.php
+++ b/web/app/themes/xrnl/doe-mee.php
@@ -10,6 +10,10 @@ get_header(); ?>
   return (object) get_field($section_id);
 } ?>
 
+<?php function insertURL($page_id) {
+  echo get_permalink(icl_object_id($page_id, 'page', true));
+} ?>
+
 <div class="join">
   <div class="bg-blue text-center text-white join-cover-image py-5" style="background: linear-gradient(rgba(0, 0, 0, 0.65), rgba(0, 0, 0, 0.45)), url('<?php the_field('join_cover_image_url'); ?>') no-repeat;">
       <h1 class="display-2 text-uppercase font-xr"><?php the_title(); ?></h1>
@@ -26,7 +30,7 @@ get_header(); ?>
       <div class="row">
         <div class="col-12 col-sm-10 col-md-8 col-lg-6 col-xl-6 mx-auto">
           <h2><?php echo($section->heading); ?></h2>
-          <p><?php echo($section->content); ?></p>
+          <?php echo($section->content); ?>
         </div>
       </div>
     </section>
@@ -40,9 +44,11 @@ get_header(); ?>
           <h2><?php echo($section->heading); ?></h2>
           <p><?php echo($section->description); ?></p>
           <div class="join-action-btns">
-            <?php foreach ($section->broadcasts as $broadcast) : ?>
-              <a class="btn btn-pink my-2" href="<?php echo($broadcast['link']); ?>"><?php echo($broadcast['label']); ?></a>
-            <?php endforeach; ?>
+            <?php if (is_array($section->broadcasts)) : ?>
+              <?php foreach ($section->broadcasts as $broadcast) : ?>
+                <a class="btn btn-pink my-2" href="<?php echo($broadcast['link']); ?>"><?php echo($broadcast['label']); ?></a>
+              <?php endforeach; ?>
+            <?php endif; ?>
           </div>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


web/app/themes/xrnl/doe-mee.php
```diff
@@ -0,0 +1,123 @@
...
+  <?php $section = getSection('groups_section'); ?>
+  <?php if ($section->enabled) : ?>
+    <section class="join-section bg-white px-2">
+      <div class="row">
+        <div class="col-12 col-sm-10 col-md-8 col-lg-6 col-xl-6 mx-auto">
+          <h2><?php echo($section->heading); ?></h2>
+          <p><?php echo($section->description); ?></p>
+          <div class="join-action-btns">
+            <?php foreach ($section->groups as $group) : ?>
+              <a class="btn btn-yellow my-2" href="/<?php echo($group['type']); ?>"><?php echo($group['button_label']); ?></a>
```
SamLubbers, line 61: Right now, `$group['type'])` always links to the Dutch page. I suggest making the `$group['type'])` custom field a URL rather than a dropdown, in order to easily link to the page of the right language. 

I also think it is a good idea to replace `href="/<?php echo($group['type']); ?>">` with `href="<?php echo($group['type']); ?>">` (without the slash) so that all URLs entered from the Wordpress admin panel are absolute URLs. Also this gives us more flexibility, maybe one day we'll want to link to a groups page only available on the global website. 

To clarify the custom field, you can change `$group['type']` to `$group['URL']` but it is not necessary.

web/app/themes/xrnl/doe-mee.php
```diff
@@ -0,0 +1,123 @@
...
+
+  <?php $section = getSection('do_more_section'); ?>
+  <?php if ($section->enabled) : ?>
+    <section class="join-section container-fluid bg-blue px-2">
+      <div class="row">
+        <div class="col-12 col-sm-10 col-md-9 col-lg-8 col-xl-8 mx-auto">
+          <h2><?php echo($section->heading); ?></h2>
+          <p><?php echo($section->description); ?></p>
+          <div class="row do-more-actions">
+            <?php foreach ($section->actions as $action) : ?>
```
SamLubbers, line 103: Like before, I recommend preceding this `foreach` with an if statement.

web/app/themes/xrnl/doe-mee.php
```diff
@@ -0,0 +1,123 @@
...
+
+  <?php $section = getSection('actions_section'); ?>
+  <?php if ($section->enabled) : ?>
+    <section class="join-section bg-blue text-white join-cover-image px-2" style="background: linear-gradient(rgba(0, 0, 0, 0.65), rgba(0, 0, 0, 0.45)), url('<?php echo($section->cover_image); ?>') no-repeat;">
+      <div class="row">
+        <div class="col-12 col-sm-10 col-md-8 col-lg-6 col-xl-6 mx-auto">
+          <h2><?php echo($section->heading); ?></h2>
+          <p><?php echo($section->description); ?></p>
+          <div class="join-action-btns">
+            <?php foreach ($section->broadcasts as $broadcast) : ?>
```
SamLubbers, line 43: I recommend preceding this `foreach` with an if statement checking if `$section->broadcasts` exists. Otherwise an error is rendered on the page.

You can reuse the code you've written for line 107: `<?php if(is_array($action['buttons'])) : ?>`

web/app/themes/xrnl/doe-mee.php
```diff
@@ -0,0 +1,123 @@
...
+
+  <?php $section = getSection('groups_section'); ?>
+  <?php if ($section->enabled) : ?>
+    <section class="join-section bg-white px-2">
+      <div class="row">
+        <div class="col-12 col-sm-10 col-md-8 col-lg-6 col-xl-6 mx-auto">
+          <h2><?php echo($section->heading); ?></h2>
+          <p><?php echo($section->description); ?></p>
+          <div class="join-action-btns">
+            <?php foreach ($section->groups as $group) : ?>
```
SamLubbers, line 60: Like before, I recommend preceding with if statement.

PR #77, opened 2020-08-21
Events REST API

jessejk89:


SamLubbers (changes requested):
Hey @jessejk89 amazing that you got the automatic event synchronization to work :fist_raised: :tada: 

Your code is very well structured, with a single function for each of the API endpoints, and it seems secure to me, since the API can only be interacted with by the script running on the same local host. I'm also very impressed by your good understanding of Wordpress (which is way better than mine :sweat_smile: ) and your use of complex SQL queries that interact with the Wordpress database.

Here are a few final points

### Pull request description

For me to do the review better, it would be useful if you could provide a description of how the code works. It would be useful to just have a simple explanation of how each of the functions works, if possible with some explanation of $query, $meta_query and other information that might not be too easy to understand by itself

### Local group field

Use `local_group` instead of `location`. Currently events are filtered by location, but soon they will be filtered by local group because the location of most events is currently online. @moritzb and @viktork are currently working on this. You can decide whether to wait until that functionality is ready, or merge this code now and incorporate the necessary changes later

### Script version control

I think the script for synchronizing the events isn't under version control at the moment, meaning that only you have access to it and no one else can improve it/use it. I don't think it's safe to have it under version control on Github, but it would be a good idea to put it on https://code.organise.earth/ If you need an account there you can message @codeadmin on Mattermost I believe

SamLubbers (commented):
Hey @jessejk89, sorry for my last comment. Only now I see that you've been committing all the code for the syncer to the fb-event-syncer repo :sweat_smile: 

Regarding my comment about the pull request description. I don't think it's necessary anymore. If you see that the script is working flawlessly on the dev website I trust that it is good enough for the live website. If I were to do an in-depth review of all php specific functions I'll probably delay for too long the release of the syncer.

Lastly, what I mention about the user of the event organiser instead of location has been merged in #130. I would merge master back into this branch, fix anything you need, and if it works well on development, then publish the code to master :ok_hand:

jessejk89:
I added some comments, I hope they clarify the code. If necessary I can explain more using a video meeting, but that might take a lot more time, like you said. 

I only use location in the filter of the events_query endpoint but it is not used by the script and not relevant anyway. And as far as I can see the changes in #130 do not touch my code.

I merged master and the code is still working fine. Only, I don't see the commits from the merge except the merge commit itself in the pull request. 

I will test the code on development as soon as possible.

SamLubbers (approved):
@jessejk89 thanks for adding the comments and great to hear that everything is still working as expected after merging the latest code from the master branch :partying_face: 

And you're right. You don't need to make any changes as a result of #130. That filter uses the `organiser` field, which your event script already fills in. :ok_hand: 

Whenever you tell me I'll merge to master and deploy the code :rocket:

PR #80, opened 2020-09-03
Template updates

franei:
This implements one small change in the footer:
- Update the link to XR UK, as they've recently moved to a new domain.

...and two small changes on Events page:

- Remove the sentence below the search results that referred visitors to Facebook or Meetup.

- Don't refer people to FB and Meetup when there are no search results; instead show a "Clear filters" button which will reload the page without filters.

image

And two small changes for the resources template:

- Wrap the library iframe in a `div` for better scrolling behaviour on iOS.
- Send a message to the resource library telling it that it's being loaded inside an iframe.

When the library gets the message it will hide its (yet to be deployed) navigation bar. 

Like this, we can show a navigation bar when visitors arrive at resources.extinctionrebellion.nl from a direct link (which happens, because resources and categories get shared by linking to them, also on our social media). Currently, visitors see no navigation at all when arriving via direct link, and there is no obvious way of getting to the main website.

### DIFF
--- a/web/app/themes/xrnl/archive-meetup_events.php
+++ b/web/app/themes/xrnl/archive-meetup_events.php
@@ -207,9 +207,13 @@ get_header(); ?>
 					</ul>
 				<?php endif ?>
 			</nav>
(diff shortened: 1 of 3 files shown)


SamLubbers (approved):
great work @franei ✊ 

The events page looks more reliable now that we don't like to Facebook and Meetup, and thank you for continuing to improve the resources page (and well spotted the old UK link 😉 )

PR #81, opened 2020-09-06
Petitions template and custom fields

reinvdwoerd:
Image Pasted at 2020-9-6 13-42
Image Pasted at 2020-9-6 13-44

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5f54069104562.json
+++ b/web/app/themes/xrnl/acf-json/group_5f54069104562.json
@@ -15,7 +15,7 @@
                 "class": "",
                 "id": ""
             },
-            "wpml_cf_preferences": 0,
+            "wpml_cf_preferences": 1,
             "return_format": "url",
             "preview_size": "full",
             "library": "all",
@@ -352,6 +352,7 @@
                         "class": "",
                         "id": ""
                     },
+                    "wpml_cf_preferences": 1,
                     "return_format": "url",
                     "preview_size": "large",
                     "library": "all",
@@ -361,8 +362,7 @@
                     "max_width": "",
                     "max_height": "",
                     "max_size": "",
-                    "mime_types": "",
-                    "wpml_cf_preferences": 0
+                    "mime_types": ""
                 }
             ]
         }
@@ -384,5 +384,5 @@
     "hide_on_screen": "",
     "active": 1,
     "description": "",
-    "modified": 1600092152
+    "modified": 1600097106
 }
--- a/web/app/themes/xrnl/acf-json/group_5f54069104562.json
+++ b/web/app/themes/xrnl/acf-json/group_5f54069104562.json
@@ -1,0 +1,897 @@
+{
+    "key": "group_5f54069104562",
+    "title": "Petition",
+    "fields": [
+        {
+            "key": "field_5f540691116a3",
+            "label": "Cover Image",
+            "name": "join_cover_image_url",
+            "type": "image",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "return_format": "url",
+            "preview_size": "full",
+            "library": "all",
+            "min_width": "",
+            "min_height": "",
+            "min_size": "",
+            "max_width": "",
+            "max_height": "",
+            "max_size": "",
+            "mime_types": "",
+            "wpml_cf_preferences": 0
+        },
+        {
+            "key": "field_5f5407a079736",
+            "label": "Form id",
+            "name": "form_id",
+            "type": "text",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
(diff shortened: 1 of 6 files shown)


reinvdwoerd:
What's still needs to be decided and made: the sections

reinvdwoerd:
Screenshot_2020-09-10 Let's Make Europe Green · Extinction Rebellion Netherlands(1)

SamLubbers (changes requested):
Hey Rein, thanks so much for making the petition page. 

I still have to review the code for the pull request, but there is still one major feature that we are missing. I'm sharing it here so that other people could also pick it up. 

When someone sign's the petition, the designers would like to show a button to donate. 

Petitionpage_2

I think it's possible to hook directly into the Action Network plugin to display the donate button. See how in point 9 of [this pull request](https://github.com/xrnl/extinction-rebellion-nl/pull/6) the displayed response was changed. Note that this will also change it for all other forms, like the one on the join page (which might not be a big problem)

Then, if people click on donate, they should be given the option to share the petition on social media

Petitionpage_3

SamLubbers:
One more thing. The API key is currently loaded form the `.env` file. This creates a bottleneck because we need to contact the server administrator every time we want to update the key. If in an emergency we need to reset the api key during an action then the petition page will also be down.

Ideally you would enter the API key into a custom field that is only accessible by admins of the website. @reinvdwoerd Do you know how to do this? If not I could give it a try.

SamLubbers:
@reinvdwoerd here are the API keys for the petition:
  * English: https://actionnetwork.org/api/v2/forms/1ec97eef-a2d3-4d2f-bbef-e228e01593a9/
  * Dutch: https://actionnetwork.org/api/v2/forms/e274d206-30ec-4443-85c0-1ce9c64746da/

I just realised a very important change! We don't only have to get the signatures from one form, but of both forms (Dutch and english) and add them up.

reinvdwoerd:
@SamLubbers thanks for feedback and research! 

I looked into the possibility of creating a custom field that's only editable by admins, it's not built into ACF but there is a filter with which I should be able to prevent the field from displaying when the user's role isn't admin, or alternatively we can register the field dynamically from PHP (example from [here](https://support.advancedcustomfields.com/forums/topic/hide-a-field-in-backend-if-non-admin/)): 

```php
add_action('acf/init', function () {
    acf_add_local_field_group(array(
        'key' => 'group_1',
        'title' => 'My Group',
        'fields' => array (
           array('key' => 'field_1', 'label' => 'Sub Title', 'name' => 'sub_title', 'type' => 'text'),
           if(is_admin()){
                array('key' => 'field_2', 'label' => 'Sub Title Admin', 'name' => 'sub_title', 'type' => 'text')
           }
        ),
        'location' => array (
            array (
                array (
                    'param' => 'post_type',
                    'operator' => '==',
                    'value' => 'post',
                ),
            ),
        ),
    ));
});
```

reinvdwoerd:
Adding up the two numbers for Dutch and English shouldn't be hard, and then I need to figure out how to show the different forms based on the current language

DaanMeijer:
> @SamLubbers thanks for feedback and research!
> 
> I looked into the possibility of creating a custom field that's only editable by admins, it's not built into ACF but there is a filter with which I should be able to prevent the field from displaying when the user's role isn't admin, or alternatively we can register the field dynamically from PHP (example from [here](https://support.advancedcustomfields.com/forums/topic/hide-a-field-in-backend-if-non-admin/)):
> 
> ```
> add_action('acf/init', function () {
>     acf_add_local_field_group(array(
>         'key' => 'group_1',
>         'title' => 'My Group',
>         'fields' => array (
>            array('key' => 'field_1', 'label' => 'Sub Title', 'name' => 'sub_title', 'type' => 'text'),
>            if(is_admin()){
>                 array('key' => 'field_2', 'label' => 'Sub Title Admin', 'name' => 'sub_title', 'type' => 'text')
>            }
>         ),
>         'location' => array (
>             array (
>                 array (
>                     'param' => 'post_type',
>                     'operator' => '==',
>                     'value' => 'post',
>                 ),
>             ),
>         ),
>     ));
> });
> ```

Actually, the way to go about this is to create an [options page](https://www.advancedcustomfields.com/resources/options-page/), and bind a field on that page. Then, it's possible to base that options page on a users permission.

reinvdwoerd:
Ah, that sounds like a better idea

SamLubbers:
Another change to make is to add smooth scrolling from the button to the petition. Like in #83

reinvdwoerd:
Yess, I left a comment about smooth scrolling in the template. We could either use [`Element.scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)

```
element.scrollIntoView({behavior: "smooth"});
```

or a library that overrides the behavior of all `#section` links with a smooth scroll.

SamLubbers (approved):
@reinvdwoerd thanks once again for setting up this page 💚 I only had to make minimal changes to make it ready for production.

I'm merging the code now (without the donate / share) follow-up so that media can start adding content. We can add this later on.

Also, because I was in a rush I made a horrible hack to translate the demands 🤮 In the upcoming days I will add this into a proper ACF options page.

PR #82, opened 2020-09-08
Overwrite CSS for the WhyDonate plugin

franei:
This defines the `whydonate` class for overwriting some of the default styling of the WhyDonate donation button, progress bar, and donation form. See [this Trello card](https://trello.com/c/LouTL1SI/70-add-whydonatenl-plugin-to-website) for the request and some before/after screenshots.

The `wd-center` class can be used for centering the button/progress bar:
```html
<div class="whydonate wd-center">
  [whydonate id="weuzk"]
</div>
```

### DIFF
--- a/web/app/themes/xrnl/assets/sass/app.scss
+++ b/web/app/themes/xrnl/assets/sass/app.scss
@@ -17,3 +17,4 @@
 @import "roles";
 @import "community";
 @import "local";
+@import "whydonate";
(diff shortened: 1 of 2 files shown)


SamLubbers:
@franei I just need to check with Marieke as she wanted to make all main CTA buttons black.

SamLubbers (changes requested):
@franei I checked with Marieke and she said that to keep the design consistent we will make all buttons black, except buttons which are on top of a dark overlay. Those buttons will be blue 👌

franei:
I've changed it the background colour. It looks like this now:
image
image

SamLubbers (approved):
Wow what a quick change 😁 Now it's all good to go 🚀

PR #83, opened 2020-09-10
Enable smooth scrolling

franei:
Addressing this [this request](https://trello.com/c/4yjGyqhQ/172-ux-smooth-scrolling) to enable smooth scrolling on the RWB page when the 'Sign up' button is clicked.

### DIFF
--- a/web/app/themes/xrnl/rwb.php
+++ b/web/app/themes/xrnl/rwb.php
@@ -168,4 +168,15 @@ get_header(); ?>
   <?php endif; ?>
 </div>
 
+<script type="text/javascript">
+  jQuery(document).ready(function() {
+    jQuery("a[href='#join']").click(function(e) {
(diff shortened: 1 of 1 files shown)


SamLubbers (approved):
@franei works very smooth 🍯 

I just realized that in the petition page we also want the same functionality. Is there a clean way to make this code work generically across pages? or is it better if it is configured per page?

franei:
Yeah there's a super clean way to do this with CSS, everywhere:
```css
html {
  scroll-behavior: smooth;
}
```

But it's not supported by Safari ☹️ 

If we want generic javascript-powered smoothness, [this page](https://css-tricks.com/snippets/jquery/smooth-scrolling/) has some good pointers.

SamLubbers:
@franei safari always ruining the day 😭 

I'll copy your code for the petition page for now

PR #84, opened 2020-09-10
SEO improvements

franei:
[ Since there are more SEO improvements on our to-do list, maybe we can keep adding to this PR as we go along ]

1. Display alt text for image on home page
- `group_5d9678d92c76e.json`: Return an array with metadata for the image, not just the URL
- `front-page.php`: Insert alt text into `` tag

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5d9678d92c76e.json
+++ b/web/app/themes/xrnl/acf-json/group_5d9678d92c76e.json
@@ -50,7 +50,7 @@
         {
             "key": "field_5d967ad7ed136",
             "label": "join cover image",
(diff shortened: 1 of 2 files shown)


SamLubbers (approved):
@franei great fix 👌 I remember Liv mentioned that we also need to add the alt tags to all other images for improved SEO. We can do that if we have spare time before the rebellion.

franei:
I think pretty much all other images are in CSS backgrounds and those can't have alt text (unlike `` elements). But I think that's fine because CSS backgrounds aren't indexed by search engines and so it won't matter for SEO.

PR #85, opened 2020-09-12
Collapsed elements in Our Movement

jessejk89:


SamLubbers (approved):
@jessejk89 this is great. Now the about us page is much easier to navigate 💯

PR #86, opened 2020-09-12
Fixed almost non-existent sign-up form margin in join page

jessejk89:


SamLubbers (approved):
@jessejk89 thanks for fixing this padding 🙌

PR #88, opened 2020-09-14
Improve blog design

jessejk89:
The CTA  is copied from the about page. It works with string translations and retrieves the cover image of the about page. There may be a better approach for this.

For this to work properly the following string translations must be added

'/en/join' -> '/join'
'Join' -> 'Doe mee'
'Join the thousands of people that are already taking real meaningful action with Extinction Rebellion. Whoever you are or however much time you have, there is a place for you within our movement. '
-> 'Sluit je aan bij de duizenden mensen die zinvolle actie ondernemen met Extinction Rebellion. Wie je ook bent of hoeveel tijd je ook hebt, er is een plek voor jou binnen onze beweging.'


SamLubbers (changes requested):
@jessejk89 thanks for this quick & necessary change 🙌  There are a few things to improve before this can be merged:

### correct permalink rendering

Instead of hardcoding url endpoints (like `/en/join`), we use the following code so that Wordpres automatically gets the right endpoint (based on the language that is selected).

```php
<?php $JoinPageURL = get_permalink(apply_filters('wpml_object_id', 7587, 'page', true)); ?>
<p><a class="btn btn-lg btn-blue" href="<?php echo $JoinPageURL ?>"><?php _e('Join', 'theme-xrnl'); ?></a></p>
```
Here `7587` is the id of the join page in the live site. 

### solid colour background

I don't think it's a good idea to get the cover image from another post (`<?php the_field('join_cover_image', 94); ?>`) because it's not obvious that those two are connected). If that custom field changes the single page will also be broken. 

The easiest way to fix this, is to use a solid colour. You can use the `bg-blue` class to use the blue background that is also used in other parts of the website.

### edit CTA from Wordpress

All text on the website should be editable from the wordpress admin panel so that media can easily change it. So instead of hardcoding the join text `Join the thousands of people... it would be great if it could be changed from a single place from within wordpress.

SamLubbers:
@jessejk89 the last point is possible with an [options page](https://www.advancedcustomfields.com/resources/options-page/) 😄 If you don't have admin rights on your local instance I can take care of this step

jessejk89:
@SamLubbers I applied the changes you suggested. I also use an ACF field for the cover image of the CTA.

The follwing ACF fields (and also a new group) need to be added from the 'Custom Blog Settings' menu item: 

- cta_cover_image -> choose a nice image

- cta_title -> MEE DOEN? | READY TO REBEL?

- cta_body -> Sluit je aan bij de duizenden mensen die zinvolle actie ondernemen met Extinction Rebellion. Wie je ook bent of hoeveel tijd je ook hebt, er is een plek voor jou binnen onze beweging. | Join the thousands of people that are already taking real meaningful action with Extinction Rebellion. Whoever you are or however much time you have, there is a place for you within our movement. 

- cta_button_text -> DOE MEE | JOIN

I do not think I have the permissions to do that. If you could do that, that would be nice, otherwise you need to give me the permissions. Either way is fine with me.

franei:
@jessejk89 Thanks so much for working on this, it's a really nice addition to the blog posts. Also, apologies that nobody has been able to look at your changes sooner. I've finally found some time, here are a few comments:

- The text fields for the title, text, and button label need to be translatable somehow.. I'm not sure what the most elegant solution is; but one way would be to have language-specific versions of the text CFs (like `cta_title_en` & `cta_title_nl`), and then make sure that the correct version gets inserted in `single.php`.
- Nice to have but not essential: an option to show or hide the CTA section. This can be useful when the code is live, but there is no content in the CFs yet (which would result in a grey block with a little blue rectangle), or just to be able to disable the CTA section temporarily, if needed.
- Maybe I'm missing something, but to me the page looked better overall with the `container` class set on the blog post, because that adds some horizontal padding for desktop-sized screen widths.
- (sorry if you already know this :)); you can include custom field groups by adding their json file in  `themes/xrnl/acf-json` to the pull request.

SamLubbers:
@franei thanks for doing the review :raised_hands: I've just assigned you as the reviewer for this pull request (instead of me). If you need help let me know :wink:

franei:
Hej @jessejk89, I just realised that the WPML plugin that we use for translation actually supports ACF option pages. So that would probably be a better way to handle it than what I suggested above :)

I'm trying it out at the moment with a different field group. If you're interested, [here's the documentation for it](https://wpml.org/documentation/related-projects/translate-sites-built-with-acf/#translate-option-pages).

jessejk89:
Hi @franei,

Thank you for reviewing. I considered your points:
>     * The text fields for the title, text, and button label need to be translatable somehow.. I'm not sure what the most elegant solution is; but one way would be to have language-specific versions of the text CFs (like `cta_title_en` & `cta_title_nl`), and then make sure that the correct version gets inserted in `single.php`.

The options where already translatable, you only need to change the language in the Wordpress top bar. I do not know if translations can be added to git by file but I mentioned the texts (that need to be added) in this chat earlier.  They can now be filled in at XRNL Settings -> Editorial.

>     * Nice to have but not essential: an option to show or hide the CTA section. This can be useful when the code is live, but there is no content in the CFs yet (which would result in a grey block with a little blue rectangle), or just to be able to disable the CTA section temporarily, if needed.

I have added that, was very simple with get_field(). For now I only show the CTA when all four fields are filled in.

>     * Maybe I'm missing something, but to me the page looked better overall with the `container` class set on the blog post, because that adds some horizontal padding for desktop-sized screen widths.

Unfortunately with the container class, the blog contents only occupy a very small region in the center on big screens. With container-fluid it resizes automatically no matter the screen size. One of the points was to widen the blog text because it was so very narrow on big screens, so I fixed that by using container-fluid and by setting the max-width to 1295px with the blog-wrapper class. It is possible that the page looks worse on some smaller desktop sizes, but for larger screens it should look better. We could increase the padding, I guess. Suggestions?

>     * (sorry if you already know this :)); you can include custom field groups by adding their json file in  `themes/xrnl/acf-json` to the pull request.

I did not know that, so thank you for pointing that out.

franei (approved):
Hey @jessejk89,

Thank you for making the CTA only show up when all fields are filled in.

You're right of course, the fields were already translatable!
Sorry to take you along on my learning experience :)

I didn't know that the blog posts were meant to be wider. All good 🙌 
And the code looks great 💯 – thank you so much!

SamLubbers (approved):

PR #89, opened 2020-09-14
Collapse demands by default in the about page

jessejk89:


SamLubbers (approved):
@jessejk89 thanks for this quick change 💯 I remember you mentioned that we should keep it open because it's so important, but Marieke said it would be much better closed so that people can more easily find the information that they are looking for.

PR #90, opened 2020-09-14
Navigation: Adjustments buttons

jessejk89:


jessejk89:
Font-size of 1.2rem doesn't seem enough on my pc to make the header fit for all sizes. It is however only a small range (Width: 1200 - 1233px). Making it 1.1rem seems to fix it but we can also consider accepting this fact or making more space another way, because smaller menu buttons do not stand out that much.

SamLubbers (changes requested):
@jessejk89 The menu looks great! The only issue I see right now is that it collapses into the hamburger way too quickly, even if there is space. I think it has to do with this massive margin:

Screenshot 2020-09-17 at 15 33 37

Could you see if you're able to remove that margin? 

Also, what changes did you made to the language selector? In my computer it looks the same as on the live site. If you could add some screenshots that would be great 🙌

jessejk89:
The collapse into the hamburger menu is fixed to 1200px due to the navbar-expand-xl class. It does not collapse automatically at the correct size, unfortunately. I could change it to navbar-expand-lg but then the breakpoint would be at 992px, which is too small. Ideally I could write custom css for this but that takes quite some time.

The change in the language selector is the position. It should now be placed after the social media icons instead of before them. 
lang_select_dff

SamLubbers (approved):
@jessejk89 thanks for the clarification! I can imagine it takes a lot of `scss` to get the navbar to work exactly as you want on different phone sizes. But even with collapsing earlier it's soooo much better with the join and donate buttons showing on mobile. Thank you so much for working last night to implement this great change 💚 

And sorry for not noticing the change of the language switcher 😓 I need to be less lazy and read the code before I ask such questions 😅

PR #91, opened 2020-09-15
Add Why Rebel page

franei:
Implements the Why Rebel page.

### Desktop
why-we-rebel_(iPad Pro)

### Mobile
why-we-rebel_(iPhone X)

### DIFF
--- a/web/app/themes/xrnl/why-rebel.php
+++ b/web/app/themes/xrnl/why-rebel.php
@@ -14,7 +14,7 @@ get_header(); ?>
 <div class="why-rebel">
 
   <?php $section = getSection('hero_section'); ?>
-  <section class="hero" style="background: linear-gradient(rgba(0, 0, 0, 0.65), rgba(0, 0, 0, 0.45)), url('<?php echo $section->image; ?>') no-repeat;">
+  <section class="hero" style="background: linear-gradient(rgba(0, 0, 0, 0.65), rgba(0, 0, 0, 0.45)), url('<?php echo $section->image; ?>')  no-repeat center center / cover;">
       <div class="container-fluid">
         <div class="row">
           <div class="col-12 col-sm-10 col-md-8 col-lg-8 col-xl-7 mx-auto">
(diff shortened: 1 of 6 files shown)


SamLubbers (changes requested):
Hey @franei I'm going to start the review now. One thing that I already see we need to change is to remove the `want to do more` section. It was in the original design, but somewhere [deep down the Mattermost channel](https://organise.earth/xr-netherlands/pl/a8wexropipfnmdf3ke6q1pdq4a) Marieke decided to remove it.

web/app/themes/xrnl/why-rebel.php
```diff
@@ -0,0 +1,165 @@
...
+  <section id="demands" class="container-fluid">
+    <div class="row">
+      <div class="col-12 col-sm-10 col-md-8 col-lg-8 col-xl-7 mx-auto">
+        <div id="demands-btn">
+          <span id="reveal-demands-btn" class="demands-toggle reveal-demands"><?php echo $labels->show_demands_text ?></span>
+          <span id="hide-demands-btn" class="demands-toggle hide-demands" style="display: none;"><?php echo $labels->hide_demands_text ?></span>
+        </div>
+        <div id="demands-list" style="display: none;">
+          <?php $about_id = (apply_filters('wpml_current_language', NULL) === 'nl') ? 94 : 421; ?>
+          <?php the_field('demands', $about_id); // Grabbing the demands from the About us page ?>
```
SamLubbers, line 95: We'll have to test this functionality when deployed. For the petition page I also had something like this working in development and then crashing when deployed 💔

SamLubbers (approved):
@franei amazing work 👏 👏 👏  The page looks exactly like the proposed design. From reading the `scss` I can tell it was quite some work to get the design right. 

Everything looks good to me 💯 I'll merge & deploy now so that Marieke can start adding content. Later we can remove the section for `want to do more`.

web/app/themes/xrnl/why-rebel.php
franei: Thanks for the review!
Sorry I missed that the want-to-do-more section was meant to be removed, I'll do that asap.
Really hope the demands are working ok 🤞

SamLubbers: @franei Yes the demands are working! 🎉 

When you remove the what to do more you can push straight to master. I don't think there is much that can go wrong 😉

franei: Oh sorry I missed your comment; I've created a new [PR](https://github.com/xrnl/extinction-rebellion-nl/pull/94) now

PR #93, opened 2020-09-16
Petition page improvements

SamLubbers:
Fixed all the bugs mentioned in [this Trello card](https://trello.com/c/xa6jEoAk).


SamLubbers:
Sorry for merging straight away without a review. These were very small bugs I quickly had to fix before we could start sharing the petition page.

PR #97, opened 2020-10-02
Avoid naming collision for local group custom post type

franei:
The `local_group` [custom post type](https://github.com/xrnl/extinction-rebellion-nl/pull/75) is causing a naming conflict with the "Local group" roles filter on the volunteer page.
In order to avoid this conflict we can just change the key under which the post type is registered with WP.
Thanks @Mananasi for figuring out the cause of the problem!


SamLubbers (approved):
Thanks for the quick fix Frank 💚

PR #101, opened 2020-11-01
Improve custom post types

### LINKED ISSUE #100: Duplicates in the vacancy roles filter
Today somebody reported an issue with the filter for vacancy roles, where some workingGroup categories appear twice, and some vacancies slip through the filter. The problem seems to be that some vacancy posts have the `&` character replaced by `&amp;` in the json array. 

Here's why I think this happens:
When someone edits a vacancy in WP using the VISUAL editor, WP replaces the ampersand character inside the json with `&amp;`. This doesn't happen when using the TEXT editor.

One way to prevent this from happening in the future is to disable the VISUAL editor, just for the `volunteer_vacancy` post type.

Adding something like this filter in `functions.php` would do the trick:
```php
    add_filter('user_can_richedit', function( $default ){
      if( get_post_type() === 'volunteer_vacancy') return false;
      return $default;
    });
```

`

image

image

#### Visual editor disabled:
image

franei:
Minor improvements for two custom post types:

1. Volunteer vacancy:
- Disable the visual editor to prevent WP from inserting `&amp;` character. Closes #100 
2. Local group:
- Add ability to associate an author with a group
- Add ability to select images from the media library (instead  of pasting in the URL)

### DIFF
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -208,7 +208,7 @@ add_action('init', function(){
     $args = array(
         'label'                 => __( 'Local Group', 'text_domain' ),
         'labels'                => $labels,
-        'supports'              => array('title'),
+        'supports'              => array('title','author'),
         'hierarchical'          => false,
         'public'                => true,
         'show_ui'               => true,
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -213,7 +213,7 @@ add_action('init', function(){
     $args = array(
         'label'                 => __( 'Local Group', 'text_domain' ),
         'labels'                => $labels,
-        'supports'              => array('title','author'),
+        'supports'              => array('title','author','revisions'),
         'hierarchical'          => false,
         'public'                => true,
         'show_ui'               => true,
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -108,6 +108,11 @@ add_action('init', function(){
     );
     register_post_type( 'volunteer_vacancy', $args );
 
+    add_filter('user_can_richedit', function( $default ){
+      if( get_post_type() === 'volunteer_vacancy') return false;
+      return $default;
+    });
+
 });
 
 add_action('init', function(){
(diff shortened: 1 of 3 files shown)


web/app/themes/xrnl/functions.php
```diff
@@ -208,7 +208,7 @@ function exclude_category( $query ) {
     $args = array(
         'label'                 => __( 'Local Group', 'text_domain' ),
         'labels'                => $labels,
-        'supports'              => array('title'),
+        'supports'              => array('title','author'),
```
SamLubbers, line 216: It's indeed useful to know the author of a local group. Maybe it's also an idea to add `editor`. That way, even if I make all local group pages, we can see who is edits that page.

And perhaps it's also useful to add `revisions` so we have a history of changes in local group pages.

SamLubbers (approved):
@franei thanks for making these changes, and for taking the time to learn about the intricacies of WordPress; even when reviewing your code I sometimes struggle to find the relevant WordPress documentation :sweat_smile:

The changes look good to me, so you can merge it any time. I only left one small suggestion in the code. 

Congratulations with surpassing our 100th pull request/issue for the website :partying_face: :birthday:

web/app/themes/xrnl/functions.php
franei: @SamLubbers thank you so much for the fast review! I should have explained my thinking better so that you don't have to dig through the documentation as well 😔

I wanted to add `author` because it enables the Author dropdown below a post. With that we can easily hand over authorship of a post to a local-group user, also if we created it originally.

In the screenshot I'm logged in as "testuser", who I've made author of the Westerpark post and who also has the [role](https://wordpress.org/support/article/roles-and-capabilities/#summary-of-roles) Author (confusingly). Like this, that user can only edit their post and nothing else on the site.

Editors and admins can still edit all posts.
Do you think that's a good setup?

The overlap in terminology around authors and editors gets pretty confusing..
Adding support for `editor` in this case enables a text editor, it doesn't change the permissions for the post :) 

Good idea to add `revisions`, I've done that 👍 

image

SamLubbers: Thank you so much for the good explanation (even with screenshots :heart_eyes:)! This is better than anything that I've ever seen on any WordPress documentation :stuck_out_tongue: 

Now I properly understand why you've added the `author` support for that role: we can have accounts who can only edit their own local group page and nothing else :ok_hand: 

I think it's fine that editors and admins can still edit it since there are only trusted people with those permissions. In the future we can look more into fine-grained permissions so that each person can only edit the part of the website relevant to them.

PR #102, opened 2020-11-06
Add ACF options sub pages

franei:
Hey @SamLubbers & @jessejk89 this is meant as a sort of proposal. If you have time, I would really appreciate your thoughts :)

Because we've recently created an options page and are planning to add more items to it, I thought it would be useful to give it some structure. I also noticed that the API key for action network is currently visible (and editable) to all users, including authors, which is maybe not quite ideal because we're about to add many more users for the local groups.

So I thought it would be worth making subpages, which thankfully is really easy.

Here I've made a settings page "XRNL Settings" with two subpages:

- Admin: only visible to admins, can be used for things like that API key
- Editorial: accessible for editors, can be used for editorial content like the CTA from [Jesse's PR](https://github.com/xrnl/extinction-rebellion-nl/pull/88), or [the demands](https://trello.com/c/i3EcsEfN/197-refactor-make-demands-a-generic-custom-field-that-can-be-imported-in-different-pages).

I've coded a template for the demands, also to check if translations are working correctly (they are).
So like this, we can now insert the demands list into any post with something like 

```php
  get_template_part('template-parts/demands-list');
```
See the included why-rebel page as an example.

Do you think this might be a useful arrangement? Any comments welcome 🌻 

image

image

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5fa311a77eb3b.json
+++ b/web/app/themes/xrnl/acf-json/group_5fa311a77eb3b.json
@@ -1,0 +1,186 @@
+{
+    "key": "group_5fa311a77eb3b",
+    "title": "Demands",
+    "fields": [
+        {
+            "key": "field_5fa31249ed2fd",
+            "label": "Demands list",
+            "name": "xrnl_demands_list",
+            "type": "repeater",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "wpml_cf_preferences": 0,
+            "collapsed": "",
+            "min": 0,
+            "max": 0,
+            "layout": "table",
+            "button_label": "Add demand",
+            "sub_fields": [
+                {
+                    "key": "field_5fa3125aed2fe",
+                    "label": "Demand",
+                    "name": "demand",
+                    "type": "group",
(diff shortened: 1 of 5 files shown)


SamLubbers (approved):
Hey @franei! thank you for exploring this and making such great changes :raised_hands: and sorry once more for not being able to review this earlier :sweat: 

I think it's really useful to have separate sub-pages for different ACF options, especially because it allows us to set different permissions on different pages. I thought that the Action Network API key was only accessible by admins and not by all users :scream: Luckily all our current website users are trustworthy and use the website properly :smile: 

I think the `editorial` and `admin` subpages are good enough for now. If the editorial page has too many custom fields at some point we can break it down further into other sub-pages. 

Also it's great that you've figured how to use template parts so that we don't have to copy paste code among different pages anymore :100: 

To finalize the pull request, I think the last thing to do would be to:
  * In `petition.php` get the Action Network API key from the correct sub page.
  * In the about, why rebel and petition page, replace the current demands with the new template part.

franei:
@SamLubbers Thank you for the review!

What do you think about merging this PR, and I create a new one for the remaining steps?

I'm thinking, like that we can first deploy this one, then I'll fill in the demands and api key, and then when we deploy the new PR the content is already there.

SamLubbers:
@franei good that you merged this first :ok_hand: once the request is approved you can merge it at any time.

PR #103, opened 2020-11-12
Improve local group post type

franei:
Updates for the local group custom post type.

These changes were requested [on discourse](https://base.extinctionrebellion.nl/t/local-group-websites/1042/12):
1. Don't require an e-mail address (because not all LGs use e-mail). In case there is no email address and also none of the social media fields are filled in, the fallback now is to show the `info@extinctionrebellion.nl` address.
2. Insert a default message above the volunteer roles.
3. Remove the broadcast channels tab & move channels to the Contacts tab.

Some additional styling changes in this PR:
1. Improve the responsive behaviour of the hero section.
2. Change the styling of the secondary navigation bar so that the active nav-link doesn't look like a button.
3. Add customisation options for the hero background color, symbol, and symbol colors. If this is a bad idea, someone please tell me :) 

Broadcast channels in the Contact tab:
utrecht

Appearance options:
image

Default message above volunteer positions:
image


SamLubbers (approved):
@franei thank these are some beauuutiful changes :sparkles: :heart_eyes: 

I love how the header blends in with the same colour of the background hero image, and that groups can now customize their page with more colours and icons.

Also I'm so glad to see that rebels have a way of sharing their feedback about the local group page and that we are incorporating their great ideas :fist_right: :fist_left: I say we but it's actually all you who is making these fantastic pages for local groups. Thank you again for putting in all this great work to empower and grow our movement :fist_raised: :green_heart:

PR #104, opened 2020-11-15
Hide overflowing text in event info

### LINKED ISSUE #79: CSS on Events page
The styling of the cards on the events page doesn't always adjust well on mobile, as sometimes the title & image are cut off on the right, and location information flows over at the bottom.

See screenshot: 
IMG_5333

franei:
This limits the event info to 2 lines and hides any overflowing text.

Closes #79. (well, partially ;) I think we can't actually prevent text on the images getting cut off, unless we want to start making sure that all images conform to a certain format)

### Before

image

### After

image

### DIFF
--- a/web/app/themes/xrnl/assets/sass/events.scss
+++ b/web/app/themes/xrnl/assets/sass/events.scss
@@ -23,6 +23,12 @@
   }
 }
 
+.event_info {
+  line-height: 1.2rem;
+  height: 2.4rem !important;
+  overflow: hidden !important;
+}
+
 .event_date {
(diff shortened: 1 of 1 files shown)


SamLubbers (approved):
@franei thanks for fixing this!!! It really took a while for one of us to find some time to add `overflow: hidden` to the `event_info` class :joy:

SamLubbers:
Oh, and it's not a big deal that the images are still cut off. If we want to make them always fit we could either increase the width of the event box and/or re-scale the image and add some borders to fill the empty space.

PR #105, opened 2020-11-23
Use ACF options

franei:
This is only a quick follow-up to #102.

The ACF options page is now on the live site, and the demands are filled in.

This just changes the `petition` and `about` templates to use the new template-part for the demands, and tells the custom field for the AN key to show up on the ACF "admin" subpage.


SamLubbers (approved):
@franei this changes looks good to go :raised_hands: 

The last ACF option change to make is to change the source of the Action Network API key right?

PR #106, opened 2020-11-23
Add ID to signup section on Join page

franei:
Addresses [this request](https://trello.com/c/ZUAHNciJ/246-html-id-toevoegen-aan-doe-mee-pagina-om-zo-te-kunnen-linken-naar-nieuwsbrief-inschrijven) to add an element ID to the sign-up section on the Join page, so that the section can be linked to directly.

The ID is translatable, so that people can use either one of these:
https://extinctionrebellion.nl/en/join#sign-up
https://extinctionrebellion.nl/doe-mee#schrijf-je-in


SamLubbers (approved):
@franei thank you for fixing this request so quickly :zap: and cool that the link is translatable :wink:

PR #107, opened 2020-11-23
Local groups index page

franei:
What's in this PR:
- A template for an index page that lists all local groups, with links to their individual pages
- Custom fields for a few settings and text snippets
- CSS

Happy to implement any feedback :)

#### Custom fields
acf-index-page

#### Index page
lg-index

### DIFF
--- a/web/app/themes/xrnl/local_group-archive.php
+++ b/web/app/themes/xrnl/local_group-archive.php
@@ -1,0 +1,131 @@
+<?php
+/*
+* Template name: Local groups
+*/
+
+get_header(); ?>
+
+<?php
+  $groups = new WP_Query([
+    'order' => 'ASC',
+    'meta_key' => 'group_name',
+    'orderby' => 'meta_value',
+    'post_type' => 'xrnl_local_group',
+    'posts_per_page' => -1
+  ]);
+  $previous_letter = '';
+?>
+
+<div class="local-groups-hero" class="container-fluid" style="--bg-image: url('<?php the_field('hero_image'); ?>')">
+  <div class="row">
+    <div class="col-11 col-sm-10 col-lg-8 mx-auto">
+      <h1 class=""><?php the_title(); ?></h1>
+      <p class="top-text"><?php the_field('hero_text') ?></p>
+    </div>
+  </div>
+</div>
+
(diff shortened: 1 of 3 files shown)


web/app/themes/xrnl/local_group-archive.php
```diff
@@ -0,0 +1,131 @@
...
+  <div class="col-11 col-sm-10 col-lg-8 mx-auto">
+
+    <div class="index-container">
+      <div class="select-region">
+        <button type="button" name="all-regions" class="active" id="show-all-btn"><?php _e('All regions', 'theme-xrnl'); ?></button>
+        <button type="button" name="noord-west" class="region-select-btn">Noord-West</button>
+        <button type="button" name="noord-oost" class="region-select-btn">Noord-Oost</button>
+        <button type="button" name="midden" class="region-select-btn">Midden</button>
+        <button type="button" name="zuid-west" class="region-select-btn">Zuid-West</button>
+        <button type="button" name="zuid-oost" class="region-select-btn">Zuid-Oost</button>
```
SamLubbers, line 43: do we also want to translate the regions?

SamLubbers (approved):
@franei thank you for getting this done :raised_hands: :raised_hands: This was one of the key missing features on the website. 

One thing I would add to the design is showing the XR logo of each local group, as most local groups have a very nice logo characteristic of their city. If you also like the idea and think it will be very easy you can add it straight away, otherwise you can add it to the Trello as a future idea and merge this pull request :ok_hand:

web/app/themes/xrnl/local_group-archive.php
franei: ah yes I wasn't sure about this.. to me it often looks weird when place names get translated, but I guess something like "Midden" is borderline.. it's as much a description as a name. Let's make them translatable then we can always change it easily in the String translation settings :)

franei:
I really like the idea of adding the logos, it will make the page look less boring. If it's ok with you i won't do it right now, because then I can help out with the Events page first (also I think it's only a handful of groups that have a logo uploaded at the moment).

SamLubbers:
@franei yes that's cool 👌

PR #108, opened 2020-11-30
Add featured event on the Events page

franei:
Implements the design below, using a new set of custom fields on the Events page.

(also, vertically centers the submit button of the event filters)

### Design:
image

### Custom fields:
image

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5fc4ba51be42e.json
+++ b/web/app/themes/xrnl/acf-json/group_5fc4ba51be42e.json
@@ -1,0 +1,209 @@
+{
+    "key": "group_5fc4ba51be42e",
+    "title": "Events",
+    "fields": [
+        {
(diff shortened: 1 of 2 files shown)


SamLubbers (approved):
@franei thanks for implementing this so quickly :raised_hands: I think the outreach team (and especially Koen) is going to be very happy :wink: And thanks also for always having such clear pull request :pray: A clear description with some screenshots makes it much easier for me to review the changes.

I think this feature is good to go: the page looks EXACTLY like the design, custom fields make it easily customizable and the on/off option is also very useful in case we don't have an event to highlight.

As a small future improvement, do you think it would be good to only show the highlighted event if no filters are selected? That way if people are looking specifically for actions or training they don't always see the big event on top.

franei:
I also find it a bit too imposing, especially on mobile it takes up quite a lot of space. Maybe the Events page could get a makeover for the new year :)

PR #109, opened 2020-12-04
Intro talk page

franei:
heya, this is the page template for the intro talk and some custom fields.
Please excuse my javascript

Any comments welcome :)

## Custom fields
image

## Page
image

## While the talk is on
image

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5fc7bc995cc35.json
+++ b/web/app/themes/xrnl/acf-json/group_5fc7bc995cc35.json
@@ -1,0 +1,423 @@
+{
+    "key": "group_5fc7bc995cc35",
+    "title": "Intro talk",
+    "fields": [
+        {
+            "key": "field_5fc7db38adf07",
+            "label": "Hero title",
+            "name": "hero_title",
+            "type": "text",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "wpml_cf_preferences": 0,
+            "default_value": "",
+            "placeholder": "",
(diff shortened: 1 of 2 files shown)


SamLubbers (commented):
Hey @franei great that you've developed this page so quickly :grin: I started the review a couple of hours ago but I haven't been able to do it because I get this error when trying to run the website on my new (Linux) work laptop:

```
Warning: require(/var/www/html/web/wp/wp-blog-header.php): failed to open stream: No such file or directory in /var/www/html/web/index.php on line 5

Fatal error: require(): Failed opening required '/var/www/html/web/wp/wp-blog-header.php' (include_path='.:/usr/local/lib/php') in /var/www/html/web/index.php on line 5
```

Since we've never had this issue before I think it has to do with my operating system (which shouldn't happen when using docker :thinking: )

The only suggestion I have so far is to have the countdown under the talk time. That way you can first see when it is and then directly how much time is left.

I know there is some pressure to have this page live, so if it works on your computer and you think it's good enough then it's okay to merge & deploy

franei:
Hmm that's odd. It looks like that file `wp-blog-header.php` doesn't exist.. but it's part of the wordpress core, so it really ought to be be there. Maybe it got corrupted somehow?

> The only suggestion I have so far is to have the countdown under the talk time. That way you can first see when it is and then directly how much time is left.

Good idea! Like this maybe?
image

> I know there is some pressure to have this page live, so if it works on your computer and you think it's good enough then it's okay to merge & deploy

I'm sure the javascript for the timer could be a lot more elegant, but as far as I can see it works ok.. 🤞

DaanMeijer:
Did you run `composer install`? That's the command responsible for putting the wordpress core in place.

PR #110, opened 2020-12-06
Fix volume dependencies

fefi42:
Hi @DaanMeijer 

People have trouble running the website without deactivating plugins. 
I think the problem originates from this instruction in the readme:
> add wordpress plugins
> curl -SL0 https://cloud.extinctionrebellion.nl/index.php/s/QCENwJwpbCoqoNB/download -o plugins.tar.gz && tar -xvf 
> plugins.tar.gz -C web/app/plugins/ && rm plugins.tar.gz

As far as I see it people install dependencies on their host system that are later used within the docker files. 
Ideally we would install dependencies directly in the container, which would solve the runs on my machine problem.

I changed the project a bit around and I think I have something that does that. 
I am not really into the website project and also don't know much about node and php. I can only help out with the docker stuff. It would be great if I could get some review on this configuration. Would this work or do I miss something important that makes it necessary to install plugins on the host?

Also I think you probably want hot loading of files. I think that should be possible by creating the volume (currently called plugins) first and then using it in the docker-compose file. Something like this:
```
volumes:
  data:
    external:
      name: extinction-rebellion-nl_plugins
```

One more thing. I recognized that the composer exits after some initial `install` command but is still part of the `docker-compose`. Is that necessary or would it be possible to just have a multistage build for the `php` service where the composer installs everything and the `php` image just copies it? This would remove the volume dependency completely and would allow you to bind the `/var/www/html` folder directly to the host. 

I am happy to help with these changes, I am just afraid that I miss the point of some of the design choices so far and at the end waste time. 

Cheers


fefi42:
Also I selected to merge on master, which is probably also not the best idea. Let me know whats your preferred way to integrate this once it is in an acceptable state.

SamLubbers:
I'm closing this PR as it's solved in #126 

Once that PR is merged we need to check if the website still runs tragically slow on macs, or if it is fixed. 

Thanks for looking into this Felix :pray:

PR #112, opened 2020-12-09
Change the slug for local groups

franei:
This changes the slug for the local groups custom post type.

After this, the permalinks for LGs will be like

```
extinctionrebellion.nl/groep/maastricht
extinctionrebellion.nl/en/groep/maastricht
```
When we deploy this we need to remember to flush the permalinks, which can be done by going to Settings > permalinks & then clicking on "save changes" (without making any changes)

We can independently decide where to put the overview page. It could be
```
extinctionrebellion.nl/lokale-groepen 
extinctionrebellion.nl/en/local-groups
```
If we change the URL for the overview page, we should remember to add a redirect from `/local` and `/en/local` to the new overview pages.

### DIFF
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -205,7 +205,7 @@ add_action('init', function(){
         'filter_items_list'     => __( 'Filter items list', 'text_domain' ),
     );
     $rewrite = array(
-        'slug'                  => 'local',
+        'slug'                  => 'groep',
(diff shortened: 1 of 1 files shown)


SamLubbers (approved):
The URLs are better like this. @franei how do you think we should keep track of all these tasks that we need to do after deploy? I think that on Trello we could have a column "to deploy" and in there each card has some tasks to do once the application is deployed.

PR #113, opened 2020-12-10
Remove the center picture on the intro talk page

franei:
The screenshot on the intro talk page is no longer needed. 
This removes it from the template and from the custom fields.

The center section then looks like this:
image


SamLubbers (approved):
looks good to me! :sparkles:

PR #117, opened 2021-01-12
Improve asset loading on local group pages

franei:
The way in which `js` and `css` assets are loaded on local group pages is not ideal at the moment. This is causing some conflicts, and a bug that currently prevents the photo gallery from being displayed.

This fixes it by loading assets the wordpress way, using the `wp_enqueue_script` and `wp_enqueue_style` functions.

Also, the javascript for controlling the behaviour of the navigation tabs and the image gallery is now in a separate file and no longer inline with the page template.

### DIFF
--- a/web/app/themes/xrnl/assets/js/xrnl-local-group-tabs.js
+++ b/web/app/themes/xrnl/assets/js/xrnl-local-group-tabs.js
@@ -1,0 +1,37 @@
+/*
+This script contains some settings for the tab controls and image gallery on
+single local group pages (single-xrnl_local_group.php).
+*/
+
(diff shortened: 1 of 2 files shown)


SamLubbers (approved):
@franei yeah I heard recently from XR Amersfoort that they failed to display the pictures. Nice that you've fixed this so quickly :raised_hands:

franei:
Well I broke it in the first place 😬 
Thanks for the super fast review!

SamLubbers:
You're very welcome 🤗 I'm trying to always do them so quickly to speed up our development⚡

PR #118, opened 2021-01-22
Local group page javascript update

franei:
#### 1. Fix a bug that prevents the mobile menu in the top navigation bar from collapsing on local group pages

The bug is caused by the LG template loading `bootstrap.js`. Bootstrap is already included in the compiled `app.js`, so it shouldn't be necessary to load it again in the first place. But for reasons that I still don't really understand, bootstrap functions like `tab()` and `carousel()` are not available with our regular setup when only `app.js` is loaded.

This now resolves the conflict in `functions.php` by first de-registering `app.js`, then loading `bootstrap.min.js`
and then the `xrnl-local-group-tabs.js` script. It only affects local group pages; all other pages work just like before.

#### 2. Get rid of owl carousel
Because:
- owl carousel is no longer in development
- bootstrap already [comes with a carousel](https://getbootstrap.com/docs/4.6/components/carousel/), so we don't need the additional javascript and css

### DIFF
--- a/web/app/themes/xrnl/assets/js/xrnl-local-group-tabs.js
+++ b/web/app/themes/xrnl/assets/js/xrnl-local-group-tabs.js
@@ -7,31 +7,31 @@ const sections = ['about', 'demands', 'actions', 'events', 'positions', 'picture
 
 function checkSectionExists(section) {
   if (! jQuery('#' + section).length) {
-    jQuery('#lg-navigation').find('a[href=\'#' + section + '\']').parent().remove();
+    jQuery('#lg-navigation')
+    .find('a[href=\'#' + section + '\']')
+    .parent()
+    .remove();
   }
 }
 
-jQuery(document).ready(function(){
+jQuery(document).ready(function($){
 
   // Remove optional sections from the navigation if they don't exist;
   // If there is only one nav-item left, remove that too.
   sections.forEach(checkSectionExists);
-  if (jQuery('#lg-navigation .nav .nav-item').length < 2){
-    jQuery('#lg-navigation .nav .nav-item').remove();
+  if ($('#lg-navigation .nav .nav-item').length < 2){
+    $('#lg-navigation .nav .nav-item').remove();
   };
 
-  // If the Over Ons section exists, make it active on page load (instead of Contact).
-  if (jQuery('#about-nav').length) {
-    jQuery('#about-nav').tab('show');
+  // If the About Us section exists, make it active on page load (instead of Contact).
+  if ($('#about-nav').length) {
+    $('#about-nav').tab('show');
   }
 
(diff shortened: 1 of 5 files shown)


SamLubbers (approved):
@franei sorry you had to deal with these cryptic Jquery issues, but I'm happy you found a solution :grin: 

And good that you used the Jquery carousel instead of using another package (I should have probably pointed that out in my first review of the local group pages :see_no_evil:)

One thing that I would still improve in the picture slide show is for the controls to be on top of the image, instead of below. That way they don't change position with different image sizes. But I'll make a separate issue for that :ok_hand:

franei:
@SamLubbers Yeah I also don't like that the controls move around when the pictures have different dimensions. The bootstrap carousel does have the controls on top –– but they still move around (it's somewhat better because the whole side of the image is clickable, but that's not very obvious).
I've added a note in the [tutorial](https://cloud.extinctionrebellion.nl/index.php/s/3eF2cnetwEHRxFk) to say that it's best when the images have the same dimensions.
We could also add some code to give them all the same height, but that's not a built-in function with the bootstrap carousel.

franei:
@SamLubbers sorry I think I misread your comment. We can put the controls above the images with just some css, then it doesn't matter how tall the pictures are.
```css

/* Make the icons green so that they are visible on white background */
    .carousel-control-prev-icon {
     background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%2314aa37' viewBox='0 0 8 8'%3E%3Cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3E%3C/svg%3E") !important;
    }
    .carousel-control-next-icon {
      background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%2314aa37' viewBox='0 0 8 8'%3E%3Cpath d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3E%3C/svg%3E") !important;
    }

/* Move the icons above the slides */
    .carousel-control-prev,
    .carousel-control-next {
      align-items: flex-start;
      top: -1.8rem;
    }

/* Align left */
    .carousel-control-prev {
      justify-content: flex-start;
    }

/* Align right */
    .carousel-control-next {
      justify-content: flex-end;
    }
```
image

SamLubbers:
@franei that's great! could you make a pull request with those changes?

franei:
Will do 👍 
Although I have to admit I'm not entirely convinced by the look of this.. maybe I can come up with something slightly prettier 😁

PR #119, opened 2021-01-22
Load svg from theme directory

franei:
The background logo on single volunteer pages is missing because apparently it's been deleted from the media library. 
This is now loading the logo from the theme directory instead.

image


SamLubbers (approved):
@franei thanks for the fix! And sorry I haven't been able to do this myself. I'm going to fix my issues running the website locally very soon :pray:

PR #120, opened 2021-01-24
Wordpress-seo plugin causes a package dependency error during docker installation

berry:
I am new to XR and XR Tech/Website. My first task is to get the XR website running locally on my Macbook. I followed the instructions in the README. 

When running `docker-compose up` I noticed some error messages in the PHP-installation and the WP website did not run. It showed 3 problems. I only show the first problem below:

```
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires wpackagist-plugin/wordpress-seo ^11.0, it is satisfiable by wpackagist-plugin/wordpress-seo[11.9] from lock repo but wpackagist-plugin/wordpress-seo[dev-trunk, 12.0, ..., 12.9.1, 13.0, ..., 13.5, 14.0, ..., 14.9, 15.0, ..., 15.6.2] from composer repo (https://wpackagist.org) has higher repository priority. The packages with higher priority do not match your constraint and are therefore not installable. See https://getcomposer.org/repoprio for details and assistance.
```

So, I followed the suggestions and set the version information of wpackagist-plugin/wordpress-seo to ^12.0. I then had to update composer.lock by running `composer update`. Now `docker-compose up` runs flawlessly and the WP website runs. 

Remarks:
- I set version of wordpress-seo to ^12.0 but there are newer version available. I choose this one to be on the conservative side.
- On inspecting the composer.lock file I noticed that WP is updated to 5.6 (current version is 5.2.2). I don't know if that causes any issues. Other packages have in composer.lock have also been updated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
     "wpackagist-plugin/wp-action-network": "^1.1",
     "ggamel/wp-sync-db": "dev-master",
     "wpackagist-plugin/events-manager-wpml": "1.2",
-    "wpackagist-plugin/wordpress-seo": "^11.0",
+    "wpackagist-plugin/wordpress-seo": "^12.0",
     "wpackagist-plugin/import-meetup-events": "^1.4",
     "wpackagist-plugin/wp-super-cache": "^1.7"
   },
--- a/composer.lock
+++ b/composer.lock
@@ -4,24 +4,21 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


SamLubbers (approved):
@berry thank you so much for fixing this issue :raised_hands: I tried it locally and your simple change also got the website working in my computer :100: 

I also looked more in depth into the composer logs and there are indeed more issues to fix. I'm also no expert in Wordpress and composer, but I think that together we can learn about it and fix any current and future issues with the website.

Congratulations with your first successful pull request :partying_face: :tada:

PR #121, opened 2021-01-31
Add citizen assembly home + faq template + acf fields

charlotte-hib:
Added 3 templates for the citizen assembly group. With the according acf. 
see design and content here: https://trello.com/c/OKiZgiFn/258-burgerberaad-page

To simplify the review here are screenshots of pages using the templates with some content:
citizen-assembly-faq.php:
citizen-assembly-faq

citizen-assembly-home.php:
citizen-assembly-home

citizen-assembly.php:
citizen-assembly


web/app/themes/xrnl/acf-json/group_6015db045b66c.json
```diff
@@ -0,0 +1,191 @@
...
+            "placeholder": "",
+            "prepend": "",
+            "append": "",
+            "maxlength": ""
+        },
+        {
+            "key": "field_6015dcb2b328a",
+            "label": "End note",
+            "name": "end_note",
+            "type": "wysiwyg",
```
charlotte-hib, line 156: I have a problem here, the editor works fine but when I tried to add content and make it bold it would not reflect in the page. See the home page screen shot.

web/app/themes/xrnl/assets/js/citizen-assembly-back-button.js
```diff
@@ -0,0 +1,25 @@
+/*
```
charlotte-hib, line 1: Better to do this in php, I will do whenever I have time

web/app/themes/xrnl/acf-json/group_6015db045b66c.json
SamLubbers: Bold text works for me :smile:

charlotte-hib: Ok locally the visual view disappeared for me for some reason but great then :)

web/app/themes/xrnl/citizen_assembly_home.php
```diff
@@ -0,0 +1,59 @@
...
+
+  <div class="container text-center py-5">
+    <div class="col-lg-12 mb-5 text-justify">
+      <?php the_content(); ?>
+    </div>
+
+    <div class="row tiles">
+      <div class="display-4 text-uppercase font-xr text-center col-12 before-text">
+        <?php the_field("tiles_before_text") ?></div>
+      <div class=" d-flex flew-row flex-wrap justify-content-center">
```
SamLubbers, line 29: I found two small design issues for the tiles on the home page:

1.  the tiles are not centred

I see that you use `justify-content-center`. But this doesn't work because the div itself is as wide as the total width of it's elements. For this class to have effect add the class `w-100` which makes the div have a width of 100%

2. on mobile screens you have to scroll quite a bit to view all links. It would be useful to have at least 2 blocks with per row for small screens. I've spent almost an half an hour messing around with CSS to get this to work, but I think I've found the right solution. 

First, instead of using `justify-content-center` use `justify-content-around`

```suggestion
      <div class=" d-flex flew-row flex-wrap justify-content-around w-100">
```

Secondly, in the `div` for an individual block. Remove all horizontal margins. These will be automatically managed by the `justify-content-around` class.

So just replace the line 31:

```html
        <div id="<?php formatElementID(get_sub_field('tile_text')); ?>" class="bg-yellow mx-autor m-4 p-2 tile-text">
```
With

```html
        <div id="<?php formatElementID(get_sub_field('tile_text')); ?>" class="bg-yellow my-2 p-2 tile-text">
```

web/app/themes/xrnl/template-parts/citizen-assembly-header.php
```diff
@@ -0,0 +1,29 @@
...
+
+<div class="join bg-yellow">
+  <div class="text-left pt-3">
+    <a id="back-button" href="<?php echo $backButtonUrl?>">
+        <h1 id="back-text"class="display-5 text-uppercase font-xr">
+            <i class="fas fa-chevron-left fa-xs px-3"></i><?php echo $backButtonText ?>
+        </h1>
+    </a>
+  </div>
+  <div class="text-center py-5 mx-auto" style="max-width:80%">
```
SamLubbers, line 26: In the header of the home page there is equal padding around the text, but in `citizen-assembly-header.php` the padding is not equal because the `<div>` used for the link adds additional vertical space. You can easily solve this with this change:

```suggestion
  <div class="text-center pb-5 pt-1 mx-auto" style="max-width:80%">
```

web/app/themes/xrnl/acf-json/group_6015db045b66c.json
```diff
@@ -0,0 +1,191 @@
...
+                    "placeholder": "",
+                    "prepend": "",
+                    "append": "",
+                    "maxlength": ""
+                },
+                {
+                    "key": "field_60168219d78aa",
+                    "label": "tile url",
+                    "name": "tile_url",
+                    "type": "page_link",
```
SamLubbers, line 94: Nice use of custom fields :fire: I didn't even know there was a custom field for selecting other pages. In most other tempaltes we just manually enter the URL

web/app/themes/xrnl/template-parts/citizen-assembly-header.php
```diff
@@ -0,0 +1,29 @@
...
+
+$current_lang = apply_filters( 'wpml_current_language', NULL );
+
+if($current_lang == 'en') {
+    $backButtonUrl = '/en/citizen-assembly';
+    $backButtonText = 'Back';
+} else {
+    $backButtonUrl = '/burgerberaad';
+    $backButtonText = 'Terug';
+}
```
SamLubbers, line 14: We've tried this approach before but for some reason on the live website it isn't able to figure out the language this way. 

The approach we follow in other pages is the following:

We use this function to get the URL based on the ID of the page.

```php
<?php function insertURL($page_id) {
  echo get_permalink(apply_filters('wpml_object_id', $page_id, 'page', true));
} ?>
```
To get the id you'll have to first create the page of the citizen's assembly page on the live website. Once you do that, you'll find in the URL the page id. The id can be of either the dutch or english page, it doesn't matter. (I think maybe even by creating a page draft you could already get the id)

Once you have the id you can get the URL of the correct language by calling the function. If your page id is `202` then you would call

```php
    <a id="back-button" href="<?php echo insertURL(202)?>">
```
With this approach, one last change that you'll have to make is using the wordpress translations for the button text (instead of hardcoding the translation in the code):

```php
<?php _e('Back', 'theme-xrnl'); ?>
```

With this code, `Back` will appear as one of the strings to translate on the WPML plugin. After adding the translation the right text will appear depending on the selected language.

SamLubbers (changes requested):
@charlotte-hib you made an amazing first contribution to the website of Extinction Rebellion NL :star: :star: :star: :star: :star:
 
The templates you've created are extremely well designed, so that almost all changes can be made from Wordpress without requiring any code changes: a `global` template to easily create new citizen assembly pages and a custom field to easily create links to those pages from the home page. 

I"ve left all my feedback as comments in the code so you can quickly make the necessary changes. In most cases I've also tried to share a concrete suggestion to solve the issue I mention :handshake:

SamLubbers (approved):
Great that you've implemented the feedback so quickly :zap: It all looks good so I'll merge your code into the master branch.

Congratulations with your first contribution to the website of Extinction Rebellion NL :partying_face:

PR #125, opened 2021-02-09
Include online events in local group pages

franei:
Hey, I've been meaning to suggest this for a while but there was also a request on MM about it recently.

It would good if the events tab on LG pages showed not only real-life events from the same place, but also
- online events that are organized by the group
- any events organized by multiple groups, where the group is co-organizer

The new filter for the main Events page will use the `Organizer name` field for those cases, which is already being filled in for most events on the website. But we don't have to wait for the filters on the main Events page to be ready before we can use it for the events tab on LG pages.

This PR adjusts the events query to include any events
- matching in physical location (using the `venue_city` field, as before)
- mentioning the group name in the `organizer_name` field.

### Before
events_before

### After

events_after

### Now matching group name to the value of this field:
image

It will match `Rotterdam`, `XR Rotterdam`, `extinction rebellion rotterdam`, `XR Utrecht & XR Rotterdam`, ...

### DIFF
--- a/web/app/themes/xrnl/single-xrnl_local_group.php
+++ b/web/app/themes/xrnl/single-xrnl_local_group.php
@@ -66,7 +66,6 @@
 
   function getLocalEvents()
   {
-    $place = get_field('place');
     $events_query = array(
       'posts_per_page' => null,
       'paged' => 1,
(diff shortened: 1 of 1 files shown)


SamLubbers (approved):
@franei what an elegant and good change! :star2: 

I hope they start using the group_name field soon so people can see events on local group pages.

PR #126, opened 2021-02-12
Install all plugins via composer

franei:
The aim of this is to install all plugins via composer so that we can have a consistent setup for developers and for the staging server.

1. Upgrade the packages required by bedrock
2. Update application.php 
3. Install WPML and ACF with [private-composer-installer](https://github.com/ffraenz/private-composer-installer)
4. Add all other missing plugins
5. Update existing plugins if possible
6. Remove redundant plugins
7. Add a composer service in docker so that composer can be used interactively for installing and removing plugins
8. Update the README

##### 1. Add the WPML & ACF license keys to the `.env` file
```
PLUGIN_ACF_KEY=_secret_key_here_
PLUGIN_WPML_KEY=_secret_key_here_
PLUGIN_WPML_ID=_secret_id_here_
```
##### 2. Pull this branch:
```sh
git pull https://github.com/franei/extinction-rebellion-nl.git plugin-updates
```
⚠️ This will break the website until the next step is complete.
##### 4. Activate plugins in the Wordpress admin


SamLubbers (commented):
@franei thank you so much for this massive and very necessary change :raised_hands: 

I've installed the website locally from scratch and it works perfectly. It installs all packages without a problem and the website looks and behaves just like the deployed version of the site. Also it's great that you've used private-composer-installer in order to hide the sensitive data of the paid WPML and ACF plugins.

The last few things that are needed before merging this branch is some final documentation:

- [x] adding the key names (without the values) to `.env.example`
- [x] add a step in the readme instructing users to ask for the value of the secret environment variables.
- [x]  remove from the readme the step for manually fetching and installing packages

I was going to make these small changes myself but it's not easy because I don't have write access for your fork. What will in such cases, and what we've started doing in the volunteer platform, is to work on upstream branches (xrnl:<branch_name>) so that we can push small changes to each other's branches. New people can still use the fork-and-pull model so that they don't need permission to make code changes.

Also, at first I forgot to set the environment variables for WPML and ACF and composer ran without a problem. Is that expected behaviour, or do we want users to see errors by enabling the `--verbose` flag?

The last thing, I see that in the description you kept a nice organized list of done and to do items. You can use the markdown checkbox notation `- [ ]` in order to make a list of checkboxes. When you're done with one item you can just check it, instead of having to edit the description.

SamLubbers:
Actually, I just spotted a few small issues. After pulling the data from production:

- the language switcher isn't available, and the dutch page is not translated. Maybe WPML isn't on by default?
- in the console I see the message `JQMIGRATE: Migrate is installed with logging active, version 3.3.2`. Is there a way for us to disable that logging? Or is this some sub-dependency that we don't have control over?

Also, I saw this unexpected request to google from our website `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://googleads.g.doubleclick.net/pagead/id. (Reason: CORS request did not succeed)`. Maybe this happens as a result of the youtube video?

SamLubbers:
I just see that at the bottom of the readme there are instructions for adding new plugins. Are those still correct?

franei:
Thank you for the comments! I will make the changes in the documentation and updated the .env.example file. Great tip about Markdown checkboxes. I kinda knew they exist but never remember to use them 🙈  
I also like the idea of being able to collaborate on a branch 🙌 

> running `composer update` not showing errors even without the environment variables added

Before pulling this branch, I would not expect an error. But after the new composer.json file has been merged, there should be an error similar to this:
```
  [FFraenz\PrivateComposerInstaller\Exception\MissingEnvException]                                
  Can't resolve placeholder {%PLUGIN_ACF_KEY}. Environment variable 'PLUGIN_ACF_KEY' is not set.
```

> language switcher not available & page not translated

That's bad :/ Could it be because WPML isn't activated yet in the Wordpress admin > plugins panel? All plugins are deactivated by default when they're first installed.

> JQMIGRATE

This ships with wordpress, it's there to check if plugin or theme functions are compatible with newer versions of jQuery that will come bundled with WP in the future. Yes, the logging can be disabled. I noticed that one alert is for a function in our own theme's `app.js` – maybe we should check if that's still needed or update it.

> Request to googleads

Yes, I'm pretty sure that originates from the embedded YT video

> readme instructions for adding new plugins still correct?

They don't work for me. When I try to run `docker-compose exec composer` with any argument, I get an error: `No container found for composer_1`. Does it work for you?
Just running `composer` is fine, but that requires having it installed locally.

SamLubbers:
@franei thanks for the changes and all your answers. We're almost there!

> But after the new composer.json file has been merged, there should be an error similar to this

True! The reason I wasn't getting an error is because I used the `-d` flag with `docker-compose up`. I'll add to #129 that if people are having issues with the installation they should run `docker logs extinction-rebellion-nl_composer_1`

> That's bad :/ Could it be because WPML isn't activated yet in the Wordpress admin > plugins panel?

I was about to check it out. But when re-installing I bumped into an error with the custom fields server :sweat_smile: 

```
composer_1  |  28/28 [============================] 100%The following exception probably indicates you have misconfigured DNS resolver(s)
composer_1  |
composer_1  |
composer_1  |   [Composer\Downloader\TransportException]
composer_1  |   Could not resolve host: connect.advancedcustomfields.com
```
I try again later.

>  They don't work for me. When I try to run docker-compose exec composer with any argument, I get an error: No container found for composer_1. Does it work for you?

I also get the same error. I looked into the error and found out that the command in the readme is incorrect. Here is what I have found out:

The syntax we need to use is `docker-compose exec <service name (from docker-compose.yml)> <command to run>`. The command from the readme tries to run the `composer` service, which is indeed declared in `docker-compose.yml`. The issue is that this service doesn't stay alive: it starts up, runs composer install, and shuts down. That's why we get the error of `No container found`. I found that [with this little hack](https://stackoverflow.com/questions/38546755/docker-compose-keep-container-running) we can keep the composer service alive. Here we have to options:

1. We keep the existing service alive after installing:

```yml
  composer:
    image: composer
    working_dir: /var/www/html
    volumes:
      - .:/var/www/html
    command: bash -c "composer install && tail -f /dev/null"
```

The disadvantage of this is that whenever you want to install a new plugin or update plugins you have to install first, and if the install fails you can't use the compose command. 

2. We create a second service that uses the composer image, but this one is just used to use composer interactively:

```yml

  composer:
    image: composer
    working_dir: /var/www/html
    volumes:
      - .:/var/www/html
    command: install

  composer-tool:
    image: composer
    working_dir: /var/www/html
    volumes:
      - .:/var/www/html
    command: tail -f /dev/null
```

With this new docker-compose file, after starting the composer-tool service, we can run `docker-compose exec composer-tool <command>` and we fix the first issue we had. 

Now comes the second issue with the commands in the readme:

the current command is `docker-compose exec composer require wpackagist-plugin/{{PLUGIN_NAME}}`, which says run the command `require` in the `composer` service. But what we actually need to run is `composer require`. What is confusing here that the service and command are both `composer`. Now with the new service it's a bit clearer

```
docker-compose exec composer-tool composer require wpackagist-plugin/{{PLUGIN_NAME}}
```
This command works for me :blush: There is probably a better way to do it with docker-compose, but this is a nice little hack for now :innocent: Let me know if this also works for you :ok_hand:

SamLubbers:
there is actually one last option to access composer, which was mentioned by Felix in #110:

> One more thing. I recognized that the composer exits after some initial install command but is still part of the docker-compose. Is that necessary or would it be possible to just have a multistage build for the php service where the composer installs everything and the php image just copies it? This would remove the volume dependency completely and would allow you to bind the /var/www/html folder directly to the host.

I think this should be possible. This will reduce the number of images & make composer available from the php service (which we can rename to server). In addition, I think it will make it easier to debug website issues. In the past, when the composer command failed, the website would still load, but it will have an obscure message like `index.php` file missing. This message also appears if someone tries to visit the website while plugins are still installing.

franei:
@SamLubbers Excellent sleuthing!
Yes, the command worked for me too after adding the `composer-tool` service 😃

SamLubbers (approved):
Hey @franei all issues are now fixed and the branch can be merged :tada: 

- This time when I tried installing the plugins with docker-compose it didn't give me the error `Could not resolve host: connect.advancedcustomfields.com` and all plugins are properly installed. 
- Now that I got the website working again locally, I discovered that indeed in my installation `WPML Multilingual CMS` wasn't activated and that's why the translations weren't working the first time around. Now the translations work perfectly. 

To really complete this pull request it would be great if you could add the code I shared about the composer-tool and adapt the instructions at the end of the readme. 

Thank you SO MUCH again for getting this to work :fist_raised: :fire: . It will save us so many installation hassles in the future, and it might even solve the issue with running the website on mac. I've left a comment on [the card](https://trello.com/c/5dB1xeZS/71-improve-website-performance) so that we can look into it in the future.

PR #127, opened 2021-02-15
Re-position image navigation for local group pages

franei:
Moving the controls to the top, as [discussed previously](https://github.com/xrnl/extinction-rebellion-nl/pull/118#issuecomment-766880665).

To my eye it looks a bit less awkward when they are in the top corners, rather than above the images. Comments welcome :)

image

Mobile, also showing highlight on hover:
image


SamLubbers (approved):
@franei great that you've changed this! And sorry I didn't see your PR before :pray: Next time if I'm not reviewing something you can notify me with a message.

These buttons do look a less awkward than having the navigation on top and they're definitely more visible than the current buttons :100:

PR #130, opened 2021-02-21
Filter events by organizer instead of location

moritzcopter:
Currently, events are filtered by location where all online events are grouped into one item - Issue with that is that we are mixing events from different local groups while users would want to see events of a local group specifically. This MR solves this by replacing the location-based filter with a filter by organizer.

Downside of this solution is that the 'Organizer' field is not always consistently used (e.g. Extinction Rebellion vs XR) and that they can be quite lengthy - see screenshot below:

image

### DIFF
--- a/web/app/themes/xrnl/archive-meetup_events.php
+++ b/web/app/themes/xrnl/archive-meetup_events.php
@@ -11,12 +11,12 @@ $param_type = stripslashes(get_query_var('type'));
 // organizer query
 $query_organizer = $param_organizer ? array(
 	'key' => 'organizer_name',
-	'value' => $param_organizer,
-	'compare' => '='
+	'value' => array($param_organizer, 'Extinction Rebellion NL'),
+	'compare' => 'IN'
 ) : array();
 
 // type query
-$query_type = $param_type ? array(
+$query_type = $param_type ? array( 
 	'key' => 'venue_address',
 	'value' => 'Online',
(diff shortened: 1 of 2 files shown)


franei:
hey @moritzcopter thank you so much for working on this! I'm not sure what we can do about the different formats in the organizer name string, other than encourage people to be consistent.
One small issue I noticed is that when selecting events by "Extinction Rebellion" almost *all* events are shown on the page (even though the number of events in the select dropdown is correct). I think this can be fixed by matching the exact phrase instead of words.
Adding
```php
'sentence' => true
```
in the `$query_organizer` array seems to do the trick (haven't done much testing though!)

moritzcopter:
Interesting, I don't see that issue with my local instance - still pushed your suggestion to be safe.

franei (approved):
Interesting that we see different behaviour 🤔 
But hey, it now all works on my instance too 😃 

I think we can definitely merge this update; just wondering if we should co-ordinate with the event synchronisers before deploying. 
At the moment for example XR Deventer shows up twice in the list, because they've entered slightly different strings in the `organizer_name` field.

moritzcopter:
Yes, good point - I have posted in the event synchronization chat: https://organise.earth/xr-netherlands/pl/tcc85eju8tghjrjafp3j18t69c
I will edit the current inconsistencies.

franei:
Awesome, thank you so much @moritzcopter 🙌

PR #131, opened 2021-02-21
Local group page updates

franei:
Some minor improvements for the local group pages:

#### 1. Add support for loading specific tabs on LG pages
In `xrnl-local-group-tabs.js`
This now checks if there is a hash parameter in the URL that matches one of the tabs, and if so, makes that tab active when the page loads. 

So LGs can use URLs like `extinctionrebellion.nl/groep/amsterdam#actions` to point directly to the Actions tab, for example.

#### 2. Small update for the custom fields:
In `group_5f1ad30f2103e.json`
- Add a link to the more detailed [instructions](https://cloud.extinctionrebellion.nl/index.php/s/3eF2cnetwEHRxFk) on NC 
- Update the help text for the events section to reflect the recent change in #125
- In the About us, Demands, and Actions sections, line breaks in the text box will now automatically insert `<p>` tags ([addressing this question](https://organise.earth/xr-netherlands/pl/ms8bxm3mh7gkjbnf3j6exiuxbc) on MM). Because of this, there are also some small formatting adjustments in the css and template file.

##### Without automatic `<p>` tags
image

##### With automatic `<p>` tags
image

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5f1ad30f2103e.json
+++ b/web/app/themes/xrnl/acf-json/group_5f1ad30f2103e.json
@@ -3,6 +3,24 @@
     "title": "Local group",
     "fields": [
         {
+            "key": "field_60327fcb79fd0",
+            "label": "Instructions",
+            "name": "",
+            "type": "message",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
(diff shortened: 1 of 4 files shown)


SamLubbers (approved):
@franei really nice that you've updated the text formatting and the description of the events. But what I'm most impressed by is the AMAZING instructions document :star_struck: The brief text and clear screenshots make it very easy to follow, and the document style is beautifully in line with the XR style guide :star2:

PR #133, opened 2021-03-05
WIP Affinity groups

charlotte-hib:
This MR is the beginning of the affinity group template. Related to this [ticket](https://trello.com/c/NfOg9J33/260-affinity-groups-page)
As I don't have the time to continue further with it now, let me explain the situation:

There is a [python script](https://github.com/martijnmartijnmartijnmartijn/rebel_management_utilities/blob/5f4e06a74cb737b12c9593e7e713a100fe94036c/rebel_management_utilities/members.py#L85) to retrieve data from the Action network, I converted it into PHP code -> see commented code (L50:L98) . It is very slow to retrieve the data so I put hardcoded data as I was working on the design.

TODO:
- check for duplicate in the $people_endpoints
- check if set_transient/get_transient is an efficient caching mechanism
- retrieve proper fields (L79). @tjallerens is the contact point for this. Here is his last message :
`
Sam asked me to share with you the names of the fields that have to be used for the website. He said you can also look in the Action Network yourself but that it might no be the best idea because is it not instantly clear how the data is saved. So, he said it would be best if I share with you the names of all the custom fields and for the dropdowns/radio buttons which values those custom fields can have. Thus we have: AG_name , AG_size , AG_risk (high , medium , low) , Email , Phone number , Municipality , AG_on_website (Yes , No) , AG_average_age , AG_description , AG_language ( Dutch , English , Dutch_English ) , AG_comments
`
- figure out which fields to display on the tiles

If you have any question, feel free to reach me directly.

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_6015db045b66c.json
+++ b/web/app/themes/xrnl/acf-json/group_6015db045b66c.json
@@ -91,7 +91,7 @@
                     "key": "field_60168219d78aa",
                     "label": "tile url",
                     "name": "tile_url",
-                    "type": "page_link",
+                    "type": "link",
                     "instructions": "",
                     "required": 1,
                     "conditional_logic": 0,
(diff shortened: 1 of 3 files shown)


SamLubbers:
I overwrote the old data because it contained sensitive information.

PR #135, opened 2021-03-10
Make event filter labels translatable

franei:
Small follow-up to #130––
String translation wasn't working without specifying the `theme-xrnl` text domain.


SamLubbers:
@franei thanks for this change :raised_hands: feel free to directly merge in the future

PR #136, opened 2021-03-16
Button analytics

SamLubbers:
This PR contains the code part of this task: https://trello.com/c/ymvxQrip

The analytics code is implemented in accordance with the specification: https://wiki.extinctionrebellion.nl/Main/Tracking

In the function I've created, I'm escaping the URL because of what's mentioned in [this SOF post](https://stackoverflow.com/questions/6768793/get-the-full-url-in-php): 

> using this code has security implications. The client can set HTTP_HOST and REQUEST_URI to any arbitrary value it wants.

To-do list:

- [x] get url with function
- [x] function with the entire tracking code syntax, taking button text as
  input.
- [x] make function globally accessible
- [x] load function in buttons that we want to track.
- [x] disable when the website is running locally.

### DIFF
--- a/web/app/themes/xrnl/class-wp-bootstrap-navwalker.php
+++ b/web/app/themes/xrnl/class-wp-bootstrap-navwalker.php
@@ -203,6 +203,11 @@ if ( ! class_exists( 'WP_Bootstrap_Navwalker' ) ) {
 			// Allow filtering of the $atts array before using it.
 			$atts = apply_filters( 'nav_menu_link_attributes', $atts, $item, $args, $depth );
 
+      // register onclick atttribute to get analytics for button clicks
+      if ($args->clicks_page_identifier) {
+        $atts['onclick'] = register_button_click($atts['title'], $args->clicks_page_identifier);
+      }
+      
 			// Build a string of html containing all the atts for the item.
 			$attributes = '';
 			foreach ( $atts as $attr => $value ) {
--- a/web/app/themes/xrnl/about.php
+++ b/web/app/themes/xrnl/about.php
@@ -24,7 +24,7 @@ get_header(); ?>
       <div class="row">
         <div class="col-12 col-lg-6 mx-auto mt-3">
           <a class="btn btn-yellow btn-lg" data-toggle="collapse" href="#demands" role="button" aria-expanded="false" aria-controls="demands" 
-onclick="<?php register_button_click('our demands') ?>">
+onclick="<?= register_button_click('our demands') ?>">
             <?php _e('OUR DEMANDS', 'theme-xrnl'); ?>
             <i class="fas fa-chevron-down"></i>
           </a>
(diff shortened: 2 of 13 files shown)


SamLubbers:
@franei this pull request is finally ready for review :tada: On top of what is mentioned above, I have:

- Added the button analytics to many more pages as this data might be useful in the future
- Tested that all buttons still work after adding the tracking code (I don't want to make you go through this process because it's a very laborious task).
  - I haven't been able to test this on Matomo because I think that Matomo only registers events from the live website. I've added a task on the Trello card to make sure we test this after deployment. 
- Right now all `onclick` tags are on a new line. I've not formatted the entire document nicely because it will make it harder to review the actual changes that I made. I will format all files automatically in a separate PR.

Lastly, I haven't had time to figure out how to add the analytics to the header and footer menu items. I think it's best if we first deploy this and then add more analytics in the future when we have some more time. Of course I've already made a [new task for it on Trello](https://trello.com/c/qwvepWZE)

SamLubbers:
Oh and one more thing! I've done some formatting at the end so the easiest to review this pull request is to look at [files changed](https://github.com/xrnl/extinction-rebellion-nl/pull/136/files) instead of individual commits.

franei (approved):
Hey @SamLubbers, thank you so much for getting this done! Great idea to write a function instead of hardcoding the onclick attribute on every page 😎 

I think you're right that we can't really test this until it's on the live site. But all looks good to me and I'm pretty sure it won't do any harm 😇 

Here's an idea for how we could add the onclick attriibutes in the header & footer menus:

1. Add something like this in `class-wp-bootstrap-navwalker.php`:

```php
if ($args->register_clicks) {
  $atts['onclick'] = register_button_click($atts['title'] . ($args->clicks_id ? '-' . $args->clicks_id : ''));
}
// e.g. somewhere around line 180
```

2. Add these two arguments where the menu is registered in `header.php`:

```php
<?php wp_nav_menu( [
    'theme_location' => 'primary',
    'depth'	         => 2,
    'container'       => '',
    'menu_class'      => 'navbar-nav',
    'register_clicks' => true, // add onclick attributes if true
    'clicks_id'       => 'header', // optional ID string to differentiate from other buttons on the same page
    'fallback_cb'     => 'WP_Bootstrap_Navwalker::fallback',
    'walker'          => new WP_Bootstrap_Navwalker(),
] ); ?>
```

... and do the same with the menus in `footer.php`.

However. This only works with changing the last line in `register_button_click()` from `echo $onclick` to `return $onclick`; and that would mean having to add `echo` in front of all other calls to `register_button_click()`, which is a bit of a pain. Maybe there is a better way, see what you think!

web/app/themes/xrnl/class-wp-bootstrap-navwalker.php
```diff
@@ -203,6 +203,11 @@ public function start_el( &$output, $item, $depth = 0, $args = array(), $id = 0
 			// Allow filtering of the $atts array before using it.
 			$atts = apply_filters( 'nav_menu_link_attributes', $atts, $item, $args, $depth );
 
+      // register onclick atttribute to get analytics for button clicks
+      if ($args->clicks_page_identifier) {
+        $atts['onclick'] = register_button_click($atts['title'], $args->clicks_page_identifier);
+      }
+
```
SamLubbers, line 210: This looks properly aligned in my computer but not here. Probably an issue of tabs & spaces. Let's leave this here for now and fix it automatically at a later point with a code formatter.

SamLubbers:
@franei Thanks for your amazing suggestion :grin: Before I had no idea about wordpress menu walkers, but with your code it was very easy to implement the analytics.

The main change is in `web/app/themes/xrnl/class-wp-bootstrap-navwalker.php` where I've some code similar to the one you suggested. Instead of defining an argument (`register_clicks`) to enable or disable click analytics, and another (`clicks_id`) to label the analytics, I have defined just one argument (`clicks_page_identifier`) that does both. In the `wp_nav_menu` of the header and footer you can see the variable that I pass to the function.

All other changes I've made simply added an `echo` call before `register_button_click()`. I used the `<=` php shortcut, so in the end we have less characters :wink: 

When you have time please have a look and see if everything is ok. (I've tested this locally and all the buttons have the correct `onclick` attribute.

SamLubbers:
@franei I just rebased with master and tested it on the development server. It works like a charm! See the attached screenshot of my behaviour in Matomo: the folder icons are pages I visit and the other icons are buttons I clicked.
image

PR #137, opened 2021-03-25
Community group update

franei:
This is a bit of a hodgepodge of bugfixes and features for the community group pages. 

1. Added support for `author` and `revision`, so that groups can own their page and track revisions.
2. Added a custom field for twitter.
3. Hero section: currently this doesn't work well because groups use very different photos and graphics. A problem with `background-size: cover` is that some essential parts of images can get cut off on some screen sizes. Another problem is that some groups use text on their cover image, which clashes with the overlaid title.

I've now 
- made the cover image a regular ``, so that it's always shown in full
- removed the overlaid title and increased the size of the main heading underneath

(Some of this is currently done with inline css on some pages 🙀)

4. Added custom fields for optional link buttons below the main content section, for things like downloads or broadcast channels.

5. Made the main content justified for better readability.

6. Added an image gallery, pretty much exactly as on the local group pages. To avoid excessive weight there is a limit of 35 pictures and the maximum file size for one picture is 1.5MB (is that reasonable?)

7. Made some minor improvements on the overview page.
- Better responsive layout, which wasn't always working well. For this I've also upgraded bootstrap from v4.3 to 4.6 because it has better support for responsive cards. (I've checked that this doesn't break anything on other pages.. it's all fine as far as I can tell)
- Removed the redundant "More information" button from cards
- Made the card image clickable

##### To do after deploying (or for testing)
- remove the inline css on the landbouw & zorgprofessionals pages
- fill in the link button fields on the landbouw and families pages

##### Screenshots
xrnl8_8890_community_xr-nl-families_(iPad Pro)

xrnl8_8890_gemeenschap_(iPad Pro)

### DIFF
--- a/web/app/themes/xrnl/community_group-archive.php
+++ b/web/app/themes/xrnl/community_group-archive.php
@@ -31,7 +31,7 @@ get_header(); ?>
 
             <?php $img_url = get_field('group_cover_image_url') ?: $default_img_url;  // Use default pic if cover image is missing ?>
             <a href="<?php the_permalink(); ?>">
-              <img src="<?php echo($img_url); ?>" class="card-img-top" alt="XRNL Community group">
+              <img src="<?php echo($img_url); ?>" class="card-img-top" alt="<?php the_title(); ?>">
             </a>
 
                <a href="<?php the_permalink(); ?>"><h5 class="card-header text-center font-xr text-uppercase bg-yellow border-0"><?php the_title(); ?></h5></a>
--- a/web/app/themes/xrnl/community_group-archive.php
+++ b/web/app/themes/xrnl/community_group-archive.php
@@ -42,6 +42,10 @@ get_header(); ?>
               </div>
 
               <div class="card-footer bg-light border-0 text-center">
+               <?php if(get_field('group_twitter_url')): ?>
+                 <a href="<?php the_field('group_twitter_url'); ?>"  target="_blank" rel="noreferrer noopener" class="btn twitter" aria-label="twitter"><i class="fab fa-2x fa-twitter"></i></a>
+               <?php endif; ?>
+
                <?php if(get_field('group_facebook_url')): ?>
                  <a href="<?php the_field('group_facebook_url'); ?>"  target="_blank" rel="noreferrer noopener" class="btn facebook" aria-label="facebook"><i class="fab fa-2x fa-facebook-f"></i></a>
                <?php endif; ?>
--- a/web/app/themes/xrnl/community_group-archive.php
+++ b/web/app/themes/xrnl/community_group-archive.php
@@ -7,7 +7,7 @@ get_header(); ?>
 
 <?php
   $groups = new WP_Query([
-    'order' => 'DESC',
+    'order' => 'ASC',
     'orderby' => 'ID',
     'post_type' => 'community_group',
     'posts_per_page' => -1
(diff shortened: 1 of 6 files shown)


web/app/themes/xrnl/community_group-archive.php
```diff
@@ -16,32 +16,30 @@
...
-               <a href="<?php the_permalink(); ?>"><h5 class="card-header font-xr text-uppercase bg-yellow border-0"><?php the_title(); ?></h5></a>
-             <?php else :?>
-              <h5 class="card-header font-xr text-uppercase bg-yellow border-0"><?php the_title(); ?></h5>
-             <?php endif; ?>
-
-              <div class="card-body bg-light text-center">
-               <?php if($show_link_to_group_page): ?>
-                 <a href="<?php the_permalink(); ?> "class="btn btn-black"><?php _e('Learn more', 'theme-xrnl'); ?></a>
+            <a href="<?php the_permalink(); ?>">
+              <img src="<?php echo($img_url); ?>" class="card-img-top" alt="XRNL Community group">
```
SamLubbers, line 34: I thin it's better to use the title of the group as alt text, just like in the individual page.
```suggestion
              <img src="<?php echo($img_url); ?>" class="card-img-top" alt="<?php the_title(); ?>">
```

SamLubbers (changes requested):
Hey @franei! This is a great hodgepodge of bugfixes and features :star2: :raised_hands: Thanks so much for taking this on! And also thank you for already listing out the things that we need to change after deployment. I've added it to the Trello card so we don't forget about it after deploying. 

Here is my feedback for each of the pages:

## single page

The picture and styling on the single page looks much better now. Also it's very nice that people can add some pictures to give new rebels an impression of the group. 

I wasn't able to test the gallery, since I still have issues with pictures on my local instance. I've added it to the list of items to test after deployment

some things that I would improve are:

- [ ] in the wordpress description of the image gallery I think it's a good idea to:
  - explain why we have a limit (for the page to load fast enough)
  - add a link to a website where they can downsize images
  - mention that for larger pictures they can upload them to Nextcloud and share a link to the folder using the buttons field. 
- [ ] place the group title above the image
- [ ] place the back button in the top left, like in most user interfaces. With so much text the back button is a bit hidden at the bottom

## overview page

For some reason the css doesn't render properly on my version of the website (see image below). Since it works well on yours, and since it will be hard to figure out what is wrong here, I recommend just testing this in the test server before deploying it to the production server. I've also added a task for this in the Trello card. I've also added a task for testing that the new version of bootstrap is automatically installed.

image

P.S. they all have the same image because my website instance still has a hard time downloading the website pictures to the right place, so I've temporarily set these myself.

web/app/themes/xrnl/community_group-archive.php
franei: @SamLubbers Thanks again for your comments, great points as always :)

I've implemented your suggestions, you can you can check them out at https://development.extinctionrebellion.nl/gemeenschap/

One thing I'm not sure about is the back button. I see what you mean about top left being the usual place for it. However I think many visitors arrive at a group page not from the overview page, but from a direct link that is shared by the groups. In that context, I wonder if telling them to go "back to see all groups" is the first thing they need to see. 
The other thing is that on mobile it looks a bit awkward to me, and that it takes up a lot of premium space. I don't have any better ideas at the moment.. see what you think.

Bootstrap was updated fine, by the way :)

image

franei: Oh and one more tiny change I made is that the main text on a group page is now left-aligned on mobile (justified didn't work well) but still justified on desktop.

SamLubbers (approved):
@franei you might not believe it, but I finally found the time to review this (and other) website pull requests :v:

All the changes you made look good! And it was really useful that you deployed on the test server so that I could have a look there, instead of testing the changes locally. We should start doing that more :+1: 

I see what you mean with the back button: on mobile it's the first thing that you read and there is some space until the title of the group. In a future version we can try making the button stand out less so that people can still find it without getting distracted by it :ok_hand:

PR #138, opened 2021-04-14
working circles information

berry:
Added custom field structure. Structure is a list of circle titles, descriptions and a link to the wiki. Also created structure_circles.php. I added a screenshot of how the /structuur page looks in development.

Screenshot_2021-04-14 Structuur · Extinction Rebellion Nederland

When deployed in production, the /structuur page needs to point to the template structure_circles. Then the content for the organisational structure can be added.

And then the second part of the Trello task can be completed: making references to the organisational structure elements.

Please review this pull request.

### DIFF
--- a/web/app/themes/xrnl/structure_circles.php
+++ b/web/app/themes/xrnl/structure_circles.php
@@ -14,35 +14,43 @@ get_header(); ?>
 <main class="structure-circles">
   <div class="container py-5">
     <div class="col-lg-12 mb-5 text-justify">
+        <header class="entry-header">
+                <h1 class="entry-title"><?php the_title(); ?></h1>
+        </header>
       <?php the_content(); ?>
     </div>
 
     <?php if( have_rows('circles') ): ?>
         <div class="col-lg-12 mb-5 text-justify">
+            <h1 class="font-xr" href="#working-groups"><?php the_field('title') ?></h1>
             <?php the_field('introduction') ?>
-        </div>
-        <div class="row">
-          <?php while ( have_rows('circles') ){ the_row();
-          $groupId = "circle-" . get_row_index();
-        ?>
-          <div class="col-12 py-2">
-            <div id="<?php echo formatElementID(get_sub_field('title')); ?>" class="mx-auto">
-              <a class="btn btn-yellow btn-lg btn-block text-left" data-toggle="collapse" href="#<?php echo $groupId; ?>"
-                role="button" aria-expanded="false" aria-controls="<?php echo $groupId; ?>">
-                <?php the_sub_field('title') ?>
-                <i class="fas fa-chevron-down float-right pt-1"></i>
-              </a>
-              <div class="text-left collapse pt-2" id="<?php echo $groupId; ?>">
-                <div class="pt-3 text-justify">
-                    <?php the_sub_field('description'); ?>
-                </div>
-                <div class="pt-3 text-justify">
-                    <a class="btn btn-lg btn-blue" href="<?php the_sub_field('link'); ?>">Wiki: <?php the_sub_field('title') ?></a>
+            <div class="row">
+              <?php while ( have_rows('circles') ){ the_row();
+              $groupId = "circle-" . formatElementID(get_sub_field('title'));
+            ?>
+              <div class="col-12 py-2">
+                <div id="<?php echo formatElementID(get_sub_field('title')); ?>" class="mx-auto">
+                  <a class="btn btn-yellow btn-lg btn-block text-left" data-toggle="collapse" href="#<?php echo $groupId; ?>"
+                    role="button" aria-expanded="false" aria-controls="<?php echo $groupId; ?>">
+                    <?php the_sub_field('title') ?>
+                    <i class="fas fa-chevron-down float-right pt-1"></i>
+                  </a>
+                  <div class="text-left collapse pt-2" id="<?php echo $groupId; ?>">
+                    <div class="pt-3 text-justify">
+                        <?php the_sub_field('description'); ?>
+                    </div>
+                    <div class="pt-3 text-justify">
+                        <?php if (have_rows('links')) : ?>
+                          <?php while ( have_rows('links') ){ the_row(); ?>
+                            <a class="btn btn-lg btn-black" href="<?php the_sub_field('link'); ?>"><?php the_sub_field('label'); ?></a>
+                          <?php } ?>
+                        <?php endif; ?>
+                    </div>
+                  </div>
                 </div>
               </div>
+              <?php } ?>
             </div>
-          </div>
-          <?php } ?>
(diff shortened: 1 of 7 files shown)


web/app/themes/xrnl/structure_circles.php
```diff
@@ -0,0 +1,51 @@
...
+      <?php the_content(); ?>
+    </div>
+
+    <?php if( have_rows('circles') ): ?>
+        <div class="col-lg-12 mb-5 text-justify">
+            <?php the_field('introduction') ?>
+        </div>
+        <div class="row">
+          <?php while ( have_rows('circles') ){ the_row();
+          $groupId = "circle-" . get_row_index();
```
SamLubbers, line 26: For the `href` of each working group, it's better to use the name of the group for 2 reasons:

1. If the order of the circles changes, the link to the groups will also change, and old links will not point to the right circle anymore
2. It's harder to remember

In the Trello task I shared some tips on how you could achieve this

If the field containing the name of the local group name is `$group`, and the subfield for the group name is `$name` then you could do the following to insert the name of the group in the `href` field. `href="<?php echo $group->name; ?>"`

the only thing to add to this code is replacing the spaces in the group name with underscores. So `Media & Communications` would become `Media_&_communications`. Another option is to only use the first word, so that the links are simpler. So `Media & Communications` would become `media`, which could then be accessed via `/structure/#media`

SamLubbers (changes requested):
@berry thank you so much for making these changes and sorry it took so long to review it :pray: 

The page looks exactly like we had envisioned when creating the task :star_struck: With the custom fields you created now it's very easy to add information about the working groups.

I'll first share some feedback on the design and then some feedback on the code.

### Design feedback

- [ ] The wiki button would look better if the colour is black, so it's consistent with the colour of other buttons on the website and because black compliments better with yellow.

Looking at the current design, there are a couple of things that could make the page even better (but which weren't described at first in the Trello task)

- [ ] Instead of just having one button for the wiki, it would be useful if working circles can add any number of buttons. That way they could link to their Mattermost channel, to a document explaining how to join that group...  For an example of how this can be done, see the template `web/app/themes/xrnl/doe-mee.php:72`, and the custom field group that is used with that templte `join page sections`
- [ ] It would be good if there is a title before the description of working groups. This title should use the `font-xr` class that other titles have, and it should be a new custom field, so that people can change the text of the title from within wordpress. And of course, the title should also have a `href=#working-groups` so that other pages can link to it.

### Code feedback

- [ ] In the custom fields it's useful to add `placeholder` text so that groups know what kind of information is expected.
    - In the `title` field, I would set the placeholder: `Title of the working group: Outreach & Training, Tech...`
    - In the `description` field I would set a placeholder of `Some information about the working group`
    - In the `link` field, in addition to having a placeholder for each of the (new) fields, I would also add a small description to the field `Here you can add any number of buttons to show up on the website`
- [ ] In the code I left one comment about improving the link to each individual group
- [ ] The location of the custom field group you've created is currently configured to `Show this field group if page = structuur`. While this works, what is even better is to link a custom field group to a template, because it's the template that reads the values from the custom fields. 

image


### Other comments

- At the end of your pull request you mentioned: 

> When deployed in production, the /structuur page needs to point to the
> template structure_circles. Then the content for the organisational structure
> can be added.

If there are things that need to be done after deployment it's best to add them to the Trello task. That way we can easily keep track of everything that still needs to be done. I've added it already to this task so that you can see an example

- In the future it's useful if you give the pull request a descriptive title.  For example, in this case it might be "Working circles on structure page".  That way we can easily find this code back in the future. It's really good if you still link the Trello task in the descritption

Thanks again for making these changes :green_heart: And let me know when you'd like me to do the final review :v:

web/app/themes/xrnl/structure_circles.php
berry: Changed the code to `$groupId = "circle-" . formatElementID(get_sub_field('title'));`. This code works. I tried leaving out `circle-` but then the collapsing does not work. Leaving out the `"circle-" .` does also result in the collapsing of the group items not working. I do not understand PHP well enough (not yet at least) to understand why.

berry (commented):
I have processed all the comments on the pull request. Please review the update on this pull request. Thank you very much in advance.

[Structuur · Extinction Rebellion Nederland.pdf](https://github.com/xrnl/extinction-rebellion-nl/files/6528268/Structuur.Extinction.Rebellion.Nederland.pdf)

web/app/themes/xrnl/structure_circles.php
SamLubbers: @berry I figured out what is wrong here! When you remove the `"circle-"` part then there are two html elements with the same id (see both elements that use `formatElementID`). Because of this, when you click on the button a second time, it doesn't only hide the description, but the div containg also the button. Keeping it like this works fine. If you want to make it even more descriptive you can make the second id `$groupId = formatElementID(get_sub_field('title')) . "-description";`

SamLubbers (changes requested):
@berry thank you for implementing all the suggestions :green_heart: It works great :sunglasses: There are only a couple of small suggestions that I made which can be fixed in a few minutes. 

This is almost going live!! :grin: :rocket:

web/app/themes/xrnl/structure_circles.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Template name: Structure - Circles
+ */
+
+get_header(); ?>
+
+<?php
+  function formatElementID($str) {
+    return strtolower(str_replace(array(' ', ' & '), '-', $str));
```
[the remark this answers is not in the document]
SamLubbers: Oh and since this function is used in multiple places, could you also fix it there? Or maybe even better, have a single version of this function in `functions.php`

web/app/themes/xrnl/structure_circles.php
```diff
@@ -0,0 +1,59 @@
...
+    <div class="col-lg-12 mb-5 text-justify">
+        <header class="entry-header">
+                <h1 class="entry-title"><?php the_title(); ?></h1>
+        </header>
+      <?php the_content(); ?>
+    </div>
+
+    <?php if( have_rows('circles') ): ?>
+        <div class="col-lg-12 mb-5 text-justify">
+            <h1 class="font-xr" href="#working-groups"><?php the_field('title') ?></h1>
```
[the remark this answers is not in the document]
berry: Fixed. I should have noticed that mistake. 😄

berry (commented):
Moved formatElementID to functions.php. Extended the function to also include ', '. Because that character combination was used on the about-page. Made sure that all formatElementID functions used on other pages are removed. Tested all pages where formatElementID was used. And also fixed the id-attribute error on the structure-circles page.

SamLubbers (approved):

PR #140, opened 2021-05-01
Disable theme and plugin updates from the WP admin

franei:
Hi @SamLubbers and @DaanMeijer 
Installing ACF and WPML with private-composer-installer seems to be working fine (#126) 🥳 
This now disables plugin and theme updates in the admin panel, like we discussed yesterday


SamLubbers (approved):
@franei I'm so glad to hear that the new plugins setup works 😁 and great that you already made this last change to complete the setup 👌

PR #146, opened 2021-07-06
Action Network form improvements

SamLubbers:
The main purpose of this change is to:

1. [adding an opt-in](https://trello.com/c/WplmXcny), so people give explicit consent for getting contacted by XR
2. [showing the field labels](https://trello.com/c/SDEqWeJ2), so people know why we ask certain data (such as phone number and municipality.

Of course, when working on this I got frustrated by the inconsistent styles so I also changed them so that there is consistent font-type, font-size, margins and paddings.

Also I removed all redundant code from this `scss` file, and documented and organised the file so that we better understand what the effect is of all that custom css.

Note that if you test this locally you won't see the opt-in option. I need to add that to the Action Network form after we deploy the new css

## Image previews

Here is how the form looks, before and after the css changes

### form with old css (before submission)

image

Screenshot from 2021-07-06 15-38-50

### form with new css (before submission)

desktop:

Screenshot from 2021-07-06 13-59-23

mobile:

Screen Shot 2021-07-06 at 13 59 45

### form with new css (after submission)

Screenshot from 2021-07-06 13-54-09

### DIFF
--- a/web/app/themes/xrnl/assets/sass/form.scss
+++ b/web/app/themes/xrnl/assets/sass/form.scss
@@ -25,17 +25,6 @@
     color: #000 !important;
   }
 
-  #can_embed_form form {
-    input,
-    select {
-      &:focus,
-      &:active {
-        // we use a pink bg, so make the box shadow dark pink
-        box-shadow: 0 0 0.2rem 0.2rem rgba(163, 33, 167, 0.25);
-      }
-    }
-  }
-
   h2 {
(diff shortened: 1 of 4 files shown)


franei (approved):
Hey @SamLubbers thank you for fixing the form! The code looks great and it's much better to have all the form-related CSS in one file. Also nice to add more comments – that makes it easier for people who want to start working on the website. Will try to remember doing that more often too 😇

PR #147, opened 2021-07-08
Disable users API endpoint

franei:
This disables the `users` endpoint from the REST API in order to prevent Wordpress from exposing usernames. See [this trello task](https://trello.com/c/v8y2OLuM/287-wordpress-exposing-usernames).


SamLubbers (approved):
@franei super nice and neat solution! (p.s. I can't believe this isn't a default feature in Wordpress :facepalm: )

PR #148, opened 2021-07-08
Custom user role for local groups

franei:
This creates a new user role for local groups with custom capabilities, as discussed in [this trello card](https://trello.com/c/4S9UUSnu/182-security-users-minimal-permissions).

1. Creates a new role `xrnl_group` & gives it permission to
- edit own posts (but not publish new posts)
- edit own pages (but not publish new pages)
- delete own posts
- upload media files

2. Adds custom capabilities to the `meetup_events` post type

3. Allows xrnl_group, editor, and admin users to
- create, edit & delete own events
- edit & delete others' events


SamLubbers (approved):
@franei amazing that you've addressed this important security risk :fist_raised: it's great that all those website accounts that manage events cannot make new pages :lock: 

Thank you for taking the time to learn how to fix this problem, and also for documenting the code so well :green_heart: :100:

PR #151, opened 2021-07-30
How we work template & custom fields

SamLubbers:
@franei people in the integration circle are planning to add a lot of relevant information to the [how we work](https://extinctionrebellion.nl/aan-de-slag-bij-xr/) page so that people can learn all they need to know to help out as a volunteer in XR.

The problem with that page is that it's one big blog of text that is very hard to navigate. Thats' why they would like to add foldout sections just like in the [structure](https://extinctionrebellion.nl/structuur) page. 

What I've done in this PR is:

#### 1. improve the custom field settings of the structure page

When using these custom fields I noticed that it was very hard to add information to the fold-out sections

Screenshot from 2021-07-30 13-59-04

Now it's much easier, with more space for the text and a `visual` option to see the rendered html.

Screenshot from 2021-07-30 13-58-31

#### 2. copy the custom fields & template of the structure page and make some slight changes

This is how the new page looks like (of course, they will add many more sections):

image

### DIFF
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -942,11 +942,13 @@ if( function_exists('acf_add_options_page') ) {
 }
 
 /*
- * Function replaces ' & ' and spaces to '-' so that
+ * Function replaces some special characters so that
  * string can be used as an ID on a HTML element.
  */
 function formatElementID($str) {
-    return strtolower(str_replace(array(', ', ' & ', ' '), '-', $str));
+    $str = str_replace(array('?', ':', ','), '', $str);
+    $str = str_replace(array(' & ',' '), '-', $str);
+    return strtolower($str);
 }
 
 function xrnl_pv_shortcode()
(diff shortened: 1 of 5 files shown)


web/app/themes/xrnl/functions.php
```diff
@@ -942,11 +942,13 @@ function update_event($data) {
...
- * Function replaces ' & ' and spaces to '-' so that
+ * Function replaces some special characters so that
  * string can be used as an ID on a HTML element.
  */
 function formatElementID($str) {
-    return strtolower(str_replace(array(', ', ' & ', ' '), '-', $str));
+    $str = str_replace(array('?', ':', ','), '', $str);
+    $str = str_replace(array(' & ',' '), '-', $str);
+    return strtolower($str);
 }
```
franei, line 952: 🙌

franei (approved):
Hey @SamLubbers, thank you for making the How we work page more organised and easier to edit. The layout of the custom fields on the Structure page is also much better. I've run it on my laptop, everything works & the code looks great! :shipit:

Thank you also for explaining so clearly what you did, that makes it very easy to review 🚀

PR #153, opened 2021-08-24
Rebellion page features

SamLubbers:
In this PR I'm adding two new sections to the rebellion page:

1. section with a form where people can sign up:

image

2. section with foldout menus where people can find all the info about the rebellion

image

### DIFF
--- a/web/app/themes/xrnl/functions.php
+++ b/web/app/themes/xrnl/functions.php
@@ -1,45 +1,46 @@
 <?PHP
-include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
-include_once(ABSPATH . 'wp-admin/includes/image.php' );
+include_once(ABSPATH . 'wp-admin/includes/plugin.php');
+include_once(ABSPATH . 'wp-admin/includes/image.php');
 
 add_theme_support("post-thumbnails");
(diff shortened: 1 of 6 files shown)


SamLubbers:
@franei I will merge straight away because they need this live by tomorrow :hot_face: but once you're back you can have a look to check everything is ok!

web/app/themes/xrnl/functions.php
```diff
@@ -1057,20 +1070,28 @@ function xrnl_add_role_caps() {
...
-        "allow_self_signed" => $dev ? true : false
+      "verify_peer" => $dev ? false : true,
+      "verify_peer_name" => $dev ? false : true,
+      "allow_self_signed" => $dev ? true : false
     ),
   );
   return stream_context_create($contextOptions);
 }
+
+/*
```
franei, line 1091: @SamLubbers I think putting this into `functions.php` is a good idea but right now it's breaking other pages like 
https://extinctionrebellion.nl/noodsituatie/ and https://extinctionrebellion.nl/groep/alkmaar/, because the `getSection` function is declared in those templates as well, and they can't be declared twice.
I'll see if there are other templates that do this, so that we can remove the function from them.

PR #155, opened 2021-08-26
Navbar update

franei:
This includes the code changes for the task [Update top menu](https://trello.com/c/2TxpyKcN/317-update-top-menu), and some of the changes that we discussed for the task [Make navigation bar easier to access](https://trello.com/c/ipqokqrF/313-make-navigation-bar-easier-to-access).

It also adds a function to change the logo (which will be useful for the donation page, because we can't use the extinction symbol there).

- Changes the default logo to the hoogwater symbol
- Adds ability to explicitly set the logo in get_header()
- Removes the socials buttons
- Removes the Doe mee button from the desktop navbar

new_desktop

new_mobile


SamLubbers (approved):
@franei I'm so happy with all these improvements :grin: such small changes that make such a big difference to the website.

It's nice that we are now using the new XR logo, and very smart that you thought of the ability to customize the logo so that we don't use it on the fundraising page. 

I will assume that most lines in `header.php` are automatically formatted and that the only thing you changed there is adding the logo in the get_header function. To make future reviews easier I will soon auto-format all website files (with [intellephense](https://intelephense.com/)) so that all styling is consistent and so that we don't have big auto-formatting changes in future PRs :sparkles:

PR #156, opened 2021-08-31
Security updates

franei:
- Update Wordpress from 5.6 to 5.8
- Update Wordpress plugins
- Remove outdated npm packages


SamLubbers (approved):
@franei the update looks good! Thanks for always taking care of upgrading & clearing the packages we use on the website :pray: :green_heart:

PR #157, opened 2021-09-12
Banner home page

SamLubbers:
@franei the rebellion coordinators wanted to have the banner for the rebellion on the website before they send out the next newsletter on Monday (in less than 12 hours). This week I've had to postpone and postpone, and I've ended up doing this now in a couple of hours. Because they want it so urgently I'll again merge & deploy straight away :roll_eyes: but I would still value your feedback when you have some time :pray:  

I've made 2 main improvements to the home page:

1. the banner for a highlighted event (in this case the rebellion event)
2. improvements to the hero section

### 1. banner

Here is how it looks on desktop and on mobile

image

image

A few things to note about this banner:

1. you can change the text & link from Wordpress, but not the colours at the moment
2. the whole banner is clickable (I chose this option because the button is very hidden to the side on desktop)
3. it's not complete: some things to improve are adding the countdown and maybe also an hourglass icon

### 2. cover

When updating the cover to match the new design that Kaj proposed I found two issues, which I fixed.

The first issue is that the overlay was too dark, and therefore it made the image stand out less. I slightly decreased the opacity of this overlay. Here are some before and after pictures. Here is a picture of both side by side (with the right one being the new one, with a less strong overlay). I'm not sure if this is right, maybe we can make the overlay even lighter, or maybe we need to make it again darker for accessibility.

Screenshot from 2021-09-13 01-11-29

Secondly, I made the cover image nice and centred. I did that by adding `background-position: 50%;` (hidden all the way at the bottom of [this commit](https://github.com/xrnl/extinction-rebellion-nl/commit/b43f0c7d060caa9c42a05a5a6a41ae481d1e9dff))

Here are some before and after pictures:

Screenshot from 2021-09-13 00-26-41

Screenshot from 2021-09-13 00-28-10

Also on mobile the image is now centred (instead of zooming on the left side)

Screenshot from 2021-09-13 00-27-05

Screenshot from 2021-09-13 00-28-02

### DIFF
--- a/web/app/themes/xrnl/acf-json/group_5d9678d92c76e.json
+++ b/web/app/themes/xrnl/acf-json/group_5d9678d92c76e.json
@@ -3,6 +3,121 @@
     "title": "Home",
     "fields": [
         {
+            "key": "field_613e806d6898d",
+            "label": "Event highlight section",
+            "name": "event_highlight_section",
+            "type": "group",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "layout": "block",
+            "wpml_cf_preferences": 0,
+            "sub_fields": [
+                {
+                    "key": "field_613e812168992",
+                    "label": "enabled",
+                    "name": "enabled",
+                    "type": "true_false",
+                    "instructions": "",
+                    "required": 0,
+                    "conditional_logic": 0,
+                    "wrapper": {
+                        "width": "",
+                        "class": "",
+                        "id": ""
+                    },
+                    "message": "",
+                    "default_value": 0,
+                    "ui": 1,
+                    "ui_on_text": "",
+                    "ui_off_text": "",
+                    "wpml_cf_preferences": 0
+                },
+                {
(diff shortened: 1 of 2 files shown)


franei:
@SamLubbers It looks really good 🙌 

The positioning of the image is great, and I think the overlay works ok like this. For better accessibility, maybe it's worth a try to increase the font size or make it bolder.

I'd maybe consider removing the DOE MEE button from the banner because 
- It already says Doe mee in "Doe mee met the klimaatrebellie"
- Actually the whole banner is clickable (good!)
- There are other DOE MEE buttons on the page that lead somewhere else
- It would save vertical space on mobile (where the button takes up an additional row)

SamLubbers:
Thanks for the quick feedback @franei I also thought that the "doe mee" button was maybe superfluous and would conflict with the other "doe mee" buttons, but I wasn't sure if people would realise that the whole thing is clickable :sweat_smile: 

I will soon remove this & also increase the font size for accessibility :sparkles:

franei:
Right, I see what you mean. To me the whole banner kind-of looks like a gigantic clickable button but not sure if other people see it like that. Maybe a hover effect would make it more explicit? (at least on desktop..)

PR #158, opened 2021-09-16
Update Wordpress to 5.8.1

franei:
Only a minor version bump but it has some more [security updates](https://wordpress.org/support/wordpress-version/version-5-8-1/).


SamLubbers (approved):
@franei LGTM :wink:

PR #159, opened 2021-09-25
Add missing dependencies for generating the theme

cdfa:
This has not been an issue so far because npm throws transitive dependencies in the same node_modules as direct dependencies.
All these dependencies were also dependencies of `laravel-mix` so they were present, but we depend on them directly so they should be added.

### DIFF
--- a/web/app/themes/xrnl/package.json
+++ b/web/app/themes/xrnl/package.json
@@ -10,14 +10,20 @@
         "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
     },
     "devDependencies": {
+        "babel-loader": "^8.2.2",
         "bootstrap": "4.6.0",
         "cross-env": "^5.1",
+        "css-loader": "^1.0.1",
         "jquery": "^3.2",
         "laravel-mix": "^4.0.7",
         "lodash": "^4.17.5",
         "popper.js": "^1.12",
+        "postcss-loader": "^3.0.0",
         "resolve-url-loader": "^2.3.1",
         "sass": "^1.15.2",
-        "sass-loader": "^7.1.0"
+        "sass-loader": "^7.1.0",
+        "style-loader": "^0.23.1",
+        "vue-template-compiler": "^2.6.14",
+        "webpack": "^4.46.0"
     }
 }


SamLubbers (commented):
@cdfa amazing that you found & fixed this issue. I have two points of feedback:

1. You had configured your PR to merge into the `develop` branch, but PRs in this repository should always merge into the `master` branch. That is because we only use `develop` for testing on the [XR NL dev website](https://development.extinctionrebellion.nl/), and we never merge from `develop` to mater. If you want to test something in the develop branch you can just force push it to that branch (`git push -f origin your_branch:develop`), and then you can deploy it to develop (you can ask me through direct message how you can do this). I already changed the target branch from `develop` to `master` so there is nothing that you need to do :wink: 
2. As you can see, this PR does not only contain your commit, but also a bunch of commits from Frank. To fix this, what you want to do is an interactive rebase with the master branch. You can edit your commit history in one of two ways:

### option 1: git (soft) reset

First you reset your branch, so that the last commit is the latest one from the master branch, and none of the new changes are committed

```
git switch master
git pull
git switch missing-theme-gen-deps
git reset master
```

Then only commit your new change

```
git add package.json
git commit -m "Add missing dependencies for generating the theme "
git push -f origin missing-theme-gen-deps
```

### option 2: git (interactive) rebase

This is a bit harder, but an interactive rebase is very useful and it's good practice 

```sh
git switch master
git pull
git switch missing-theme-gen-deps
git rebase -i master
```
After these commands you'll get an editor where you can pick and choose which commits to keep and which ones to drop. To delete Frank's commits, you can either remove the lines of the commits that you don't want to keep, or replace in those commits `keep` for `drop`. Once you're done you can save the interactive rebase file. (if your editor opens the interactive rebase in vim, you can delete lines with `dd`, move down and up with `j` and `k`, and save the file with `:wq` + enter). Once you've completed the interactive rebase you can just force push your changes.

---

I probably shared here :point_up: a lot of stuff that you already know how to do. But I'd rather share too much than too little in case it takes me again a while to reply :innocent:

cdfa:
Oh weird that I suddenly have Frank's commits on my branch. Was there a force push to `develop`?
Anyway, I did already know most of the stuff you shared about git. It was nice of you to add it, but I don't think it would be necessary in the future :).

Hope everything is alright like this!

franei:
Oops, yes sorry I have a bad habit of force-pushing to the develop branch 🥺 

This is probably a dumb question– but isn't it how it's supposed to be, that laravel-mix & webpack deal with all the dependencies for building the theme?

cdfa:
No problem @franei .

> This is probably a dumb question– but isn't it how it's supposed to be, that laravel-mix & webpack deal with all the dependencies for building the theme?

That's not a dumb question at all. It really depends on what you think is more "right". `laravel-mix` relies on incidental `npm` behaviour to control the dependencies, but it does make sense to have them under it's control. The "official" way to do what `laravel-mix` is doing is to declare these dependencies as peer-dependencies, but that might require people to install some dependencies they don't really use, which isn't great either. The only real solution would be to split up the package, but I don't know to what extent this is possible since it's called `laravel-*mix*`.

franei:
Thanks for explaining @cdfa, that's interesting. I need to learn more about this!

SamLubbers (approved):
@cdfa thanks for rebasing. The PR is good to go! :rocket: 

I've also just invited you to the website team. That way you can make a branch in the main repository, instead of having to make one on your fork

PR #160, opened 2021-09-29
updated time of the website meeting

SamLubbers:
@franei very small but necessary change :v: I left it to every other thursday because we'll probably revert to that after the rebellion


franei:
Well spotted :)
Yes every other week seems good to me as well 👍

PR #161, opened 2021-09-30
Remove donate button from footer partial

hansrochat:
Removed donate button from the footer as per [Trello card](https://trello.com/c/wag0q06m/318-remove-all-mentions-from-donations-except-in-navigation-bar).

remove-donate-btn-footer

### DIFF
--- a/web/app/themes/xrnl/footer.php
+++ b/web/app/themes/xrnl/footer.php
@@ -1,82 +1,87 @@
-    </main>
-    <footer class="bg-black pt-5">
-        <div class="container">
-            <div class="row">
-                <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


SamLubbers (approved):
@hansrochat thanks for this nice quick change :raised_hands: the only improvement I see is that next time you make separate commits for the auto-formatting and for the code change. That way it's easier to see what has actually changed. I'll approve & merge now. congratulations with your first code contribution to the website :partying_face: 

@tomootes since you're working on the footer you might want to rebase from master before continuing.

hansrochat:
@SamLubbers Code formatting was due to a Sublime Text extension, which formats according to WordPress coding standards. Usually comes in quite handy, but not in this case. I will use a different IDE (PHPStorm) for future code edits.

PR #162, opened 2021-09-30
Demand 0

SamLubbers:
New design for the list of demands, including the new demand 0. I'll first list the design decisions I made and which you might want to comment on. At the bottom you can find screenshots for how it looks on Desktop and on mobile. To review the changes it's easiest to review each of the commits one by one.

#### Design decisions

- make background colour the colour of the climate justice demand & make a distinct colour for our main 3 demands. I hope that this illustrates that demand 0 is the base for our 3 demands.
- display demands side by side on Desktop (instead of on top of each other like before). I like this from the XR global website as it saves some nice vertical space and makes full use of the screen width. Also this allows people to make a screenshot of the website and use this in the slides for talks / trainings etc.
- always display the demands on the about page, instead of showing them if user clicks on "our demands". I think it's important to always show the demands because it shows what we are working towards (regardless of what the media or people say about us). Also this makes the design much nicer because now the demands can occupy the full width of the page.
- I removed the demands from the emergency page because it didn't work with the new design and I don't think it's necessary to also have it on that page.
- I added a big title and a small pre-title (above the title), with exactly the responsive styles of the global page. I hope we can reuse this for other new sections.

#### Screenshots

image

Screenshot 2021-09-30 at 13-02-30 Over ons · Extinction Rebellion Nederland

(edit) I see that in this second screenshot I forgot to adjust the colour of the link at the end. This will be ok once we deploy it and I add the content through wordpress.

### DIFF
--- a/web/app/themes/xrnl/about.php
+++ b/web/app/themes/xrnl/about.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Template name: About
  */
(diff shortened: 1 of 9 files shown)


franei (approved):
@SamLubbers I think this looks great, and also shows really nicely that the demand 0 is both work in progress and underlying to the other 3 🎯 

One thing I noticed is that the text underneath the demands is white on my laptop, but gray in the screenshots, not sure if that's intentional? (sorry if we already talked about this yesterday). Either way white looks fine to me too :)

Completely agree that it's better to have the demands on the page immediately 🙌 

One more comment about the text, as demand 0 is meant to be the underlying principle of demands 1, 2 and 3, should it still say "Doe dit op een rechtvaardige manier" in the second one and "rechtvaardige transitie" in the third? 

Not sure why I'm asking you this.. it's probably not something we should just change 😇

SamLubbers:
@franei thanks for the review :raised_hands:

The text underneath the demands is indeed white on your laptop because I've added the css through wordpress, rather than directly in the page. In the website meeting we decided that we should actually integrate that into the code. I'll do that and then deploy :rocket: 

And about the text, I think that only once demand 0 is accepted the movement will decide whether to remove those bits from the demands 2 & 3

PR #174, opened 2022-04-14
Events api changes

jessejk89:
Added a way to allow more than one machine to access the events api by using the optional variable EVENT_REQUEST_EXTRA_IPS=IP1,IP2,IP3 in the .env file. 

The methods for listing events ( /wp-json/events_api/v1/events) and retrieving a single event ( /wp-json/events_api/v1/events/<id>) were recoded to rely on the wordpress api instead of error-prone sql statements. This fixes some bugs with duplicate events and makes the code a lot simpler.

All events api code was moved to a seperate file.

An attempt was made to make the events API work with the Events Manager plugin, as can be seen in some commits. These changes have been reverted.


web/app/themes/xrnl/events-api.php
```diff
@@ -0,0 +1,401 @@
...
+      '_cover_bytes_hash' => $data['_cover_bytes_hash']
+    ),
+  );
+
+  // if no author is specified, it defaults to the first available user
+  // so specifying a user is better
+  $syncerUser = get_user_by('login', 'fb-events-syncer');
+  if($syncerUser) {
+    $post['post_author'] = $syncerUser->id;
+  }
```
SamLubbers, line 183: I think the comment doesn't match the code, right?

What I see in the code is that it will always assign the `fb-events-syncer` user, whereas in the comment it says that it sets the author passed through the API call

SamLubbers, line 183: I think the comment doesn't match the code, right?

What I see in the code is that it will always assign the `fb-events-syncer` user, whereas in the comment it says that it sets the author passed through the API call

web/app/themes/xrnl/events-api.php
```diff
@@ -0,0 +1,401 @@
...
+  // When a picture_url is supplied, the following code creates a new attachment
+  // for the picture and uses the id of that attachment as _thumbnail_id
+  // Supplying an existing picture id as _thumbnail_id is enough to add that picture to the event
+  if($data['picture_url'] != NULL) {
+    $attach_id = uploadPicture($data['picture_url'], $data['title']);
+    $post['meta_input']['_thumbnail_id'] = $attach_id;
+  }
+
+  // create the post
+  $err = wp_insert_post($post, true);
```
SamLubbers, line 194: `$err` is not the right variable name, because it can also not be an error no? If I'm correct, if there are no errors the value is the post id

SamLubbers, line 194: `$err` is not the right variable name, because it can also not be an error no? If I'm correct, if there are no errors the value is the post id

web/app/themes/xrnl/events-api.php
```diff
@@ -0,0 +1,401 @@
...
+
+  if(!is_wp_error($err)) {
+    // if no error occured, turn the post into a meetup event
+    wp_set_object_terms($err, $data['category'], 'meetup_category');
+  }
```
SamLubbers, line 199: The comment says that here is where it turns it into a meetup event, but didn't that already happen when you defined `    'post_type' => 'meetup_events',`? What I think you're doing here is just setting the categories for this event, right? If so, I have another question: could you set these categories before the post is created? or does it need to happen after?

SamLubbers, line 199: The comment says that here is where it turns it into a meetup event, but didn't that already happen when you defined `    'post_type' => 'meetup_events',`? What I think you're doing here is just setting the categories for this event, right? If so, I have another question: could you set these categories before the post is created? or does it need to happen after?

web/app/themes/xrnl/events-api.php
```diff
@@ -0,0 +1,401 @@
...
+  return $err;
+}
+
+/**
+ * Upload a picture as attachment and return it's id
+ * @param $url The web url of the picture
+ * @param $title The title of the picture. A random suffix will be added to it
+ */
+function uploadPicture($url, $title) {
```
SamLubbers, line 208: If the same picture is uploaded, is there a way to use the picture that is already in the media library, instead of uploading the same one again? For example, by comparing file names.

 I'm asking because we have so many duplicate pictures in our library, and I think it comes from the event syncer.

SamLubbers, line 208: If the same picture is uploaded, is there a way to use the picture that is already in the media library, instead of uploading the same one again? For example, by comparing file names.

 I'm asking because we have so many duplicate pictures in our library, and I think it comes from the event syncer.

web/app/themes/xrnl/events-api.php
```diff
@@ -0,0 +1,401 @@
...
+      $filename = get_attached_file( $thumbnail_id );
+      //echo "Thumbnail already exists, updating " . $filename
+      updatePicture($data['picture_url'], $filename);
+    } else {
+      $attach_id = uploadPicture($data['picture_url'], $original_post->post_title);
+      $post['meta_input']['_thumbnail_id'] = $attach_id;
+    }
+  }
+
+  // for all possible event fields: check if they need updating
```
SamLubbers, line 294: I see that this code below is almost the same as the code in `insert_event_meetup`. Is there a way to refactor it so that they use the same code? I think that would be great as setting all these fields is a lot of lines of code and it would be good if we only define it once.

SamLubbers, line 294: I see that this code is almost the same as the code in `insert_event_meetup`. Is there a way to refactor it so that they use the same code? I think that would be great as setting all these fields is a lot of lines of code and it would be good if we only define it once.

SamLubbers (approved):
Hey @jessejk89 thanks for this great change :star2: really nice that you refactored the SQL statements into the wordpress API. Now it's soo much easier to understand, and therefore less error prone and more maintainable. also great that you moved all of the API code to a separate file. We should do that more of the code in functions.php

#### improvements

I see there is some automatic formatting. This makes the review slightly more time consuming as there are lines that Github says has changed, but are actually just autoformatted differently. Next time it would be useful if you first make separate PRs for just the changes and just the formatting. (p.s. at some point we want to auto-format all php files and use decide on a consistent automatic formatter, so we don't have this problem in the future)

Besides this, I've left all other suggestions in the code. Most of them are refactoring suggestions and not necessary code changes to make the application work better. So once you've changed what you want you can merge your code into the main branch :partying_face:

web/app/themes/xrnl/functions.php
```diff
@@ -850,11 +381,21 @@ function update_event($data)
...
-        __('You are not allowed to access this resource'),
-        array('status' => 401)
-      );
+      $ips = preg_split ("/,/", EVENT_REQUEST_EXTRA_IPS);
+      if(!in_array($_SERVER['REMOTE_ADDR'], $ips)) {
+        // return new WP_Error(
+        //   'forbidden',
+        //   __('Your machine with IP address ' . $_SERVER['REMOTE_ADDR'] . ' is not allowed to access this resource' ),
+        //   array('status' => 401)
+        // );
```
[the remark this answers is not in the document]
jessejk89: This is useful for developers so they can easily find out the IP address from the perspective of the website's docker container. The event synchronizing does not work when that IP address is not whitelisted. This could be put elsewhere perhaps, for example in the readme of the python script, but this way, it's just a matter of uncommenting. For now I will put an explanation in comment. Do you have a suggestion what to do with it?

web/app/themes/xrnl/events-api.php
jessejk89: Yes, you're right that it doesn't say what the code does, however, it does explain the default behaviour of wordpress when making a post and thus the need to set the post author explicitly. So wordpress defaults to Aaron when we don't set it, because it's the first user probably. But I will make the comment clearer.

web/app/themes/xrnl/events-api.php
jessejk89: It only sets the category. Don't remember why that comment was there, but indeed it's false, so I will remove it. As far as I know setting it must be done after creating a post.

web/app/themes/xrnl/events-api.php
jessejk89: This is a tricky one because Facebook regenerates the picture links every know and then. At that point it's hard to tell if it's actually changed or not. A hash of the bytes of the picture and comparing them might work but is beyond the scope of this pull request. I will make an issue for this in the fb-events-syncer github page.

jessejk89: Correction, I am already comparing byte array hashes from the old and new pictures. I am, however, probably not reusing the images of recurring events, so that can probably be improved. Will be continued.

